PR #1, opened 2016-03-18
Linter fixes

martingartonft:
I only renamed stuff, so in theory nothing could possibly break, but I DIDN'T TEST IT.


Nesze:
+1

notifications_push.go
```diff
@@ -59,33 +59,33 @@ func main() {
...
+		http.HandleFunc("/__health", hc.healthcheck())
+		http.HandleFunc("/__gtg", hc.gtg)
 
 		go func() {
 			err := http.ListenAndServe(":8080", nil)
 			errorLogger.Println(err)
 		}()
 
-		notificationsApp.consumeMessages()
+		app.consumeMessages()
```
kovacshuni: app inside and app outside?

kovacshuni:
+1

notifications_push.go
martingartonft: This will work ok but I agree might be confusing because the variable name is shadowed.

martingartonft:
As you guys are working on this still, do you mind picking this up and merging when it's convenient please?

Nesze:
It's still in progress. But we can merge this right now.

PR #2, opened 2016-03-18
instead of spawning go routines use non-blocking select block for bro…

Nesze:
…adcasting messages

### DIFF
--- a/event_dispatcher.go
+++ b/event_dispatcher.go
@@ -90,12 +90,21 @@ func (d eventDispatcher) distributeEvents() {
 		select {
 		case msg := <-d.incoming:
 			for sub := range d.subscribers {
-				go func(sub chan string) { sub <- msg }(sub)
+				select {
+				case sub <- msg:
+				default:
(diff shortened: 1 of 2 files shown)


event_dispatcher.go
```diff
@@ -90,12 +90,21 @@ func (d eventDispatcher) distributeEvents() {
...
 			}
 			resetTimer(heartbeat)
 		case <-heartbeat.C:
 			for sub := range d.subscribers {
-				go func(sub chan string) { sub <- heartbeatMsg }(sub)
+				select {
+				case sub <- heartbeatMsg:
+				default:
+					warnLogger.Println("Subscriber lagging behind when sending heartbeat.")
+					d.removeSubscriber <- sub
```
Nesze: What do you think about this?

martingartonft: lgtm
nice one.

martingartonft: that should mean that a blocking consumer can never waste resources in our app. instead they will end up dropping messages, and that's their own fault.

PR #4, opened 2016-03-18
Deadlock fix

martingartonft:
Fix what looks like a deadlock.  I admit I haven't see it or even tried to reproduce it, but it's obviously there.

PR raised mainly for discussion.

### DIFF
--- a/event_dispatcher.go
+++ b/event_dispatcher.go
@@ -103,7 +103,6 @@ func (d eventDispatcher) distributeEvents() {
 				case sub <- heartbeatMsg:
 				default:
 					warnLogger.Println("Subscriber lagging behind when sending heartbeat.")
-					d.removeSubscriber <- sub
 				}
 			}
 			heartbeat = newTimer()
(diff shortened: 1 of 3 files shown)


healthcheck.go
```diff
@@ -71,7 +71,7 @@ func (h *healthcheck) checkMessageQueueProxyReachable(address string, topic stri
 		warnLogger.Printf("Could not connect to proxy: %v", err.Error())
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
Nesze: why to wrap it?

martingartonft: Ah. I didn't intend to push this in the same pr. never mind.

This was to silence a linter by ensuring that it knows that we really don't care about the return value.

I'm happy to revert this bit?  Arguably not worth the ugliness?

To see what I mean, run gometalinter without this change

event_dispatcher.go
```diff
@@ -103,7 +103,6 @@ func (d eventDispatcher) distributeEvents() {
 				case sub <- heartbeatMsg:
 				default:
 					warnLogger.Println("Subscriber lagging behind when sending heartbeat.")
-					d.removeSubscriber <- sub
```
Nesze: I didn't get it at first sight. Deadlock happens when we remove more than one subscriber.

Nesze:
+1

PR #7, opened 2016-03-21
process msgs sequantially --> better chance of in-order msg delivery

Nesze:
Note: It still does not guarantee in-order delivery.


martingartonft:
lgtm. much better because it put the problem where it belongs (in this app) and it can be better solved here than in the gonsumer lib anyway.

PR #22, opened 2016-06-16
Save recent notifications

Nesze:
...and some structural refactoring

New endpoint is accessible at `/notifications` path. PAM will be looking at this endpoint to check for SLA failures.


kovacshuni:
checks out. +1

PR #24, opened 2016-06-27
notifications endpoint looks like original UP content notifications

kovacshuni:
I’m modifying notifications-push so that it has an endpoint `/content/notifications` that is almost exactly like the original UP notifications endpoint, the format at least, exact. It doesn’t have the same pagination because after the first page, it sends to you ?empty=true, which responds with a notifications page with [] notifications. This is for PAM to be able to check upon push.. the same way as for original notif.


Nesze:
+1

kovacshuni:
i'll write some tests

PR #25, opened 2016-07-07
trying to fix PAM regarding push monitoring

kovacshuni:


notifications.go
```diff
@@ -1,5 +1,7 @@
 package main
 
+const ChangeType = "http://www.ft.com/thing/ThingChangeType/"
```
Nesze: Should this be exported (I know, this is not a lib, but still) ?

handler.go
```diff
@@ -12,9 +12,10 @@ const evPrefix = "data: "
 const errMsgPrefix = "Serving /notifications request: [%v]"
 
 type handler struct {
-	dispatcher         *eventDispatcher
-	notificationsCache queue
-	apiBaseUrl         string
+	dispatcher           *eventDispatcher
+	notificationsCache   queue
+	apiBaseUrl           string
+	notificationsBaseUrl string
```
Nesze: Do we need this explicitly? If yes, could we come up with a better naming, like `internalBaseUrl` or something else?

If no, you can extract the `links` creation into a separate function in the `notifications` handler

kovacshuni: Renamed.

notifications.go
kovacshuni: lowercased :)

app.go
```diff
@@ -97,7 +97,7 @@ func main() {
 		infoLogger.Printf("Config: [\n\tconsumerAddrs: [%v]\n\tconsumerGroupID: [%v]\n\ttopic: [%v]\n\tconsumerAutoCommitEnable: [%v]\n\tapiBaseURL: [%v]\n\tnotifications_capacity: [%v]\n]", *consumerAddrs, *consumerGroupID, *topic, *consumerAutoCommitEnable, *apiBaseURL, *nCap)
 
 		notificationsCache := newCircularBuffer(*nCap)
-		h := handler{dispatcher, notificationsCache, *apiBaseURL}
+		h := handler{dispatcher, notificationsCache, *apiBaseURL, *apiBaseURL + "/__notifications-push"}
```
Nesze: Just a suggestion, extracting this into a builder function, would simplify things, as you would not need to pass the last parameter here, nor in the tests.

queue.go
```diff
@@ -2,33 +2,46 @@ package main
...
 type queue interface {
-	enqueue(i item)
-	dequeue() item
-	items() []item
+	enqueue(*notificationUPP)
+	dequeue() *notificationUPP
+	items() []*notificationUPP
 }
 
 type circularBuffer struct {
```
Nesze: (technically, it's not a circularBuffer anymore)

queue.go
```diff
@@ -2,33 +2,46 @@ package main
...
 
 func newCircularBuffer(capacity int) queue {
-	return &circularBuffer{&sync.Mutex{}, make([]item, 0, capacity)}
+	return &circularBuffer{&sync.Mutex{}, make([]*notificationUPP, 0, capacity), capacity}
 }
 
-func (cb *circularBuffer) enqueue(i item) {
+func (cb *circularBuffer) enqueue(n *notificationUPP) {
+	wasRemoved := true
+	for wasRemoved {
```
Nesze: do you need to nested fors here?

Nesze:
Please run `go fmt`, and some linters, `golint` or even `gometalinter`

queue.go
```diff
@@ -2,33 +2,46 @@ package main
...
 }
 
-func (cb *circularBuffer) enqueue(i item) {
+func (cb *arrayUniqueue) enqueue(n *notificationUPP) {
+	wasRemoved := true
+	for wasRemoved {
+		wasRemoved = false
+		for i, e := range cb.buffer {
+			if (e.ID == n.ID && e.Type == n.Type) {
+				cb.buffer = append(cb.buffer[:i], cb.buffer[i+1:]...)
```
Nesze: potential `slice bounds out of range` panic

https://play.golang.org/p/750O8iw3Go

app.go
kovacshuni: done

queue.go
kovacshuni: discussed, remained.

queue.go
kovacshuni: renamed

Nesze:
👍

PR #26, opened 2016-07-11
Queue discards elements with same ID and same type which are _older_.

kovacshuni:


queue.go
```diff
@@ -13,14 +16,27 @@ func newUnique(capacity int) uniqueue {
 }
 
 func (cb *uniqueue) enqueue(n *notificationUPP) {
-	wasRemoved := true
-	for wasRemoved {
-		wasRemoved = false
-		for i, e := range cb.buffer {
-			if e.ID == n.ID && e.Type == n.Type {
+	if n.LastModified == "" {
```
tosan88: is it worth to add a test for this case too? :-?

tosan88:
:+1:

queue.go
```diff
@@ -13,14 +16,27 @@ func newUnique(capacity int) uniqueue {
...
+	}
+	for i, e := range cb.buffer {
+		if e.ID == n.ID && e.Type == n.Type {
+			eDate, err := time.Parse(time.RFC3339Nano, e.LastModified)
+			if err != nil {
+				warnLogger.Printf("Notification in cache has malformed date: %v", n)
+				return
+			}
+			if eDate.Before(newestRawDate) {
 				cb.buffer = append(cb.buffer[:i], cb.buffer[i+1:]...)
```
Nesze: It does not belong to this PR, but is there a potential race condition?

Nesze:
👍

PR #30, opened 2016-11-07
Add support for list notifications

izzyblues:
This PR includes the following changes:
- Notification push can be configured to notify changes in content or lists by setting the `notification_resource` CLI parameter;
- Introduction of the `monitor` flag in order to enable PAM to consume directly from the push stream;
- General code refactoring to improve readability;
- Introduction of logrus as logging API
- Consumption of message batches in order to maintain consistent the sequence of notifications.
- `notifications` endpoint renamed as `__history`
- Increased unit test coverage of code statements


dispatcher/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+type Dispatcher interface {
+	Start()
+	Stop()
+	Send(notification ...Notification)
+	Subscribers() []Subscriber
+	Registrator
+}
+
+// Registrator :smirk:
+type Registrator interface {
```
khatton-ft: The English word would be "Registrar".

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,131 @@
...
+	Register(subscriber Subscriber)
+	Close(subscriber Subscriber)
+}
+
+// NewDispatcher creates and returns a new dispatcher
+func NewDispatcher(delay time.Duration, history History) Dispatcher {
+	return &dispatcher{
+		delay:       delay,
+		inbound:     make(chan Notification),
+		subscribers: map[Subscriber]bool{},
```
khatton-ft: What's the purpose of making this a map rather than just []Subscriber?

resources/healthchecks.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/Financial-Times/go-fthealth"
```
khatton-ft: I believe we're supposed to use go-fthealth/v1a

app.go
```diff
@@ -1,34 +1,40 @@
...
-	notificationsCache  *uniqueue
-	delay               int
+	log.SetFormatter(f)
 }
 
 func main() {
-	app := cli.App("notifications-push", "Proactively notifies subscribers about new content publishes/modifications.")
+	app := cli.App("notifications-push", "Proactively notifies subscribers about new content or lists publishes/modifications.")
+	resource := app.String(cli.StringOpt{
+		Name:   "notifications_resourse",
```
izzyblues: Fixed

app.go
```diff
@@ -1,34 +1,40 @@
...
-	consumerConfig      *queueConsumer.QueueConfig
-	notificationBuilder notificationBuilder
-	notificationsCache  *uniqueue
-	delay               int
+	log.SetFormatter(f)
 }
 
 func main() {
-	app := cli.App("notifications-push", "Proactively notifies subscribers about new content publishes/modifications.")
+	app := cli.App("notifications-push", "Proactively notifies subscribers about new content or lists publishes/modifications.")
```
izzyblues: Fixed

app.go
```diff
@@ -62,7 +68,7 @@ func main() {
 	apiBaseURL := app.String(cli.StringOpt{
 		Name:   "api_base_url",
 		Value:  "http://api.ft.com",
-		Desc:   "The API base URL where the content is accessible",
+		Desc:   "The API base URL where the content and lists are accessible",
```
izzyblues: Done

consumer/mapper.go
```diff
@@ -0,0 +1,35 @@
...
+// NotificationMapper maps CmsPublicationEvents to Notifications
+type NotificationMapper struct {
+	APIBaseURL string
+	Resource   string
+}
+
+// MapNotification maps the given event to a new notification.
+func (n NotificationMapper) MapNotification(event PublicationEvent, transactionID string) (dispatcher.Notification, error) {
+	if event.UUID == "" {
+		return dispatcher.Notification{}, errors.New("CMS publication event does not contain a UUID")
```
izzyblues: In this case, Go doesn't allow you to return nil because the method does not return a pointer to a notification.
As far as I understood, this is the best practice in Go in order to avoid what in Java are called "null pointer exceptions".

dispatcher/dispatcher.go
izzyblues: I blame @peteclark-ft here... :P

peteclark-ft: This was a cheeky coco joke from me 😳

dispatcher/dispatcher.go
izzyblues: It is more efficient for deleting subscribers and it avoids additional bugs in the code by searching and deleting an element in a slice.
With a map, adding and deleting have O(1) time complexity.
With a slice, adding has O(1) time complexity, but deleting has O(n) time complexity.
I am quite sure that HashSet is implemented in Java in the same way.

resources/healthchecks.go
izzyblues: Yes, I am going to update the API

dispatcher/dispatcher.go
khatton-ft: In that case I believe it's idiomatic to use map[Subscriber]struct{} as the type.

PR #35, opened 2016-12-14
parse UUID from ContentURI instead of taking if directly from the message

Nesze:
I'm not entirely happy with this, but at least it's now consistent with the Ingester's way of parsing the  UUID. 

One less redundant field in CmsPublicationEvents, and let's review it once the mapper work is completed.

Edit/Context: the mappers, bar the list-mapper, do not put the UUID field on the msg.


Nesze:
Just spotted something...

PR #38, opened 2017-02-01
Added discard of publish events from publish carousel

izzyblues:


consumer/publish_event.go
```diff
@@ -19,6 +19,15 @@ func (msg NotificationQueueMessage) HasSynthTransactionID() bool {
 	return strings.HasPrefix(tid, "SYNTH")
 }
 
+// HasCarouselTransactionID checks if the message is genearted by the publish carousel
+func (msg NotificationQueueMessage) HasCarouselTransactionID() bool {
+	tid := msg.TransactionID()
+	matches, _ := regexp.MatchString(carouselTransactionIDRegExp, tid)
```
peteclark-ft: I would recommend changing this to:

```
var carouselTransactionIDRegExp = regexp.MustCompile(`^(tid_[\S]+)_carousel_[\d]{10}.*$`)

...

carouselTransactionIDRegExp.MatchString(tid)
```

This saves a compile step for that regex on every message.

peteclark-ft:
Just that one compiled regex thing - otherwise LGTM.

PR #39, opened 2017-02-08
Logging when closing connection

kovacshuni:


resources/push.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Financial-Times/notifications-push/dispatcher"
 	log "github.com/Sirupsen/logrus"
+	"reflect"
```
peteclark-ft: What's this used for?

kovacshuni: sorry, mistake.

PR #45, opened 2017-07-07
Add api key validation

davidbalazs93:
As part of productionize Notifications Push work, we are introducing validation for apiKeys via Mashery call.
Please see https://sites.google.com/a/ft.com/universal-publishing/documentation/story-specs/productionising-notifications-push-api for architecture diagram and specifications.

Changes in this PR:
 - added vendoring
 - added HTTP call to Mashery to validate apiKey
 - updated Dockerfile

### DIFF
--- a/app.go
+++ b/app.go
@@ -25,8 +25,7 @@ import (
 const heartbeatPeriod = 30 * time.Second
 
 func init() {
-	f := &log.TextFormatter{
-		FullTimestamp:   true,
+	f := &log.JSONFormatter{
 		TimestampFormat: time.RFC3339Nano,
 	}
 
--- a/app.go
+++ b/app.go
@@ -19,6 +19,7 @@ import (
 	"github.com/Financial-Times/service-status-go/httphandlers"
 	"github.com/jawher/mow.cli"
 	"fmt"
+	"net"
 )
 
 const heartbeatPeriod = 30 * time.Second
@@ -144,10 +145,10 @@ func main() {
 		}
 
 		queueHandler := consumer.NewMessageQueueHandler(whitelistR, mapper, dispatcher)
-		httpClient := &http.Client{}
-		consumer := queueConsumer.NewBatchedConsumer(consumerConfig, queueHandler.HandleMessage, httpClient)
+		hc := getResilientClient()
+		consumer := queueConsumer.NewBatchedConsumer(consumerConfig, queueHandler.HandleMessage, hc)
 		masheryApiKeyValidationUrl := fmt.Sprintf("%s/%s", *apiBaseURL, *apiKeyValidationEndpoint)
-		go server(":" + strconv.Itoa(*port), *resource, dispatcher, history, consumerConfig, masheryApiKeyValidationUrl, httpClient)
+		go server(":" + strconv.Itoa(*port), *resource, dispatcher, history, consumerConfig, masheryApiKeyValidationUrl, hc)
 
 		pushService := newPushService(dispatcher, consumer)
 		pushService.start()
@@ -178,4 +179,19 @@ func server(listen string, resource string, dispatcher dispatcher.Dispatcher, hi
 
 	err := http.ListenAndServe(listen, nil)
 	log.Fatal(err)
+}
+
+func getResilientClient() *http.Client {
+	tr := &http.Transport{
+		MaxIdleConnsPerHost: 32,
+		Dial: (&net.Dialer{
+			Timeout:   10 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).Dial,
+	}
+	c := &http.Client{
+		Transport: tr,
+		Timeout:   time.Duration(10 * time.Second),
+	}
+	return c
 }
--- a/app.go
+++ b/app.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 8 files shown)


app.go
```diff
@@ -144,10 +145,10 @@ func main() {
 		}
 
 		queueHandler := consumer.NewMessageQueueHandler(whitelistR, mapper, dispatcher)
-		httpClient := &http.Client{}
-		consumer := queueConsumer.NewBatchedConsumer(consumerConfig, queueHandler.HandleMessage, httpClient)
+		hc := getResilientClient()
```
davidbalazs93: lgtm, only a small obs. wouldn't it be more readable to have httpClient instead of hc?

tosan88: Sure, I would have preferred an interface, but later on decided to not use as it would have been a bigger change

davidbalazs93: :+1:

resources/push.go
```diff
@@ -8,17 +8,39 @@ import (
 
 	"github.com/Financial-Times/notifications-push/dispatcher"
 	log "github.com/Sirupsen/logrus"
+	"io"
+	"io/ioutil"
 )
 
+const (
+	ApiKeyHeaderField = "X-Api-Key"
```
danielcopaciu: Why are these constants public?

resources/push.go
```diff
@@ -73,3 +95,50 @@ func getClientAddr(r *http.Request) string {
...
+	log.WithField("url", req.URL.String()).WithField("apiKeyFirstChars", apiKeyFirstChars).Info("Calling Mashery to validate api key")
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		log.WithField("url", req.URL.String()).WithError(err).Error("Cannot send request to Mashery")
+		http.Error(w, "Request to validate api key failed", http.StatusInternalServerError)
+		return false
+	}
+	defer func() {
+		io.Copy(ioutil.Discard, resp.Body)
```
danielcopaciu: Why is this necessary?

resources/push.go
```diff
@@ -73,3 +95,50 @@ func getClientAddr(r *http.Request) string {
...
+		http.Error(w, "Invalid api key", http.StatusUnauthorized)
+		return false
+	}
+
+	log.WithField("url", req.URL.String()).Errorf("Received unexpected status code from Mashery: %d", respStatusCode)
+	http.Error(w, "Request to validate api key returned an unexpected response", http.StatusServiceUnavailable)
+	return false
+}
+
+func isApiKeyFirstFourCharsLoggable(providedApiKey string) bool {
```
danielcopaciu: Do we really need a separate function for this check?

app.go
```diff
@@ -172,3 +179,18 @@ func server(listen string, resource string, dispatcher dispatcher.Dispatcher, hi
 	err := http.ListenAndServe(listen, nil)
 	log.Fatal(err)
 }
+
+func getResilientClient() *http.Client {
+	tr := &http.Transport{
```
tosan88: Indeed, it was a mistake (first it was _actually_ resilient, but later on decided to not use as it would have been a bigger change but didn't rename the method). I'll fix this as well

resources/push.go
tosan88: I wanted to be clear what the check is actually about. I'm happy to change that but I'd prefer this way rather then commenting on the spot

resources/push.go
tosan88: AFAIK this is needed for reusing HTTP connections, see http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

resources/push.go
```diff
@@ -73,3 +95,50 @@ func getClientAddr(r *http.Request) string {
...
+
+func validApiKey(w http.ResponseWriter, providedApiKey string, masheryApiKeyValidationURL string, httpClient *http.Client) bool {
+	req, err := http.NewRequest("GET", masheryApiKeyValidationURL, nil)
+	if err != nil {
+		log.WithField("url", masheryApiKeyValidationURL).WithError(err).Error("Invalid URL for api key validation")
+		http.Error(w, "Invalid URL", http.StatusInternalServerError)
+		return false
+	}
+
+	req.Header.Set(ApiKeyHeaderField, providedApiKey)
```
tosan88: True, will add a check

resources/push.go
tosan88: No reason, will change that

resources/push.go
```diff
@@ -73,3 +95,50 @@ func getClientAddr(r *http.Request) string {
 	}
 	return r.RemoteAddr
 }
+
+func validApiKey(w http.ResponseWriter, providedApiKey string, masheryApiKeyValidationURL string, httpClient *http.Client) bool {
```
kovacshuni: `isValidApiKey`

kovacshuni: could you move this function into a separate file, for the validation to be abstracted away, make it a separate responsibility entity.

resources/push.go
tosan88: Done

resources/push.go
tosan88: Done

resources/push.go
tosan88: Changed

app.go
tosan88: Fixed

resources/push.go
tosan88: Fair enough. Changed

kovacshuni:
i would like a separate entity for the validator like

`func (v validator) validate() {}`

so it can be inserted, injected, mocked, like in java. If you'd want to test the function that is using the validation function you'd have to set up everything for them. Otherwise you could just mock this component and mask away the validator's functionality /and the other way around.

Maybe in the next release.

PR #46, opened 2017-07-14
Remove kafka proxy and connect directly to Kafka

ioanaciont:
- Uses the internal library: https://github.com/Financial-Times/kafka-client-go
- A possible concern for these changes: the underlying library always returns `true` for Kafka healthcheck, even if it's not healthy.


ioanaciont:
[Notifications Push - talk directly to Kafka](https://trello.com/c/GyPf31VO/14-notifications-push-talk-directly-to-kafka)

consumer/consume.go
```diff
@@ -28,42 +28,42 @@ func NewMessageQueueHandler(whitelist *regexp.Regexp, mapper NotificationMapper,
...
+	notification, err := qHandler.mapper.MapNotification(pubEvent, msg.TransactionID())
+	if err != nil {
+		log.WithField("transaction_id", msg.TransactionID()).WithField("msg", string(msg.Body)).WithError(err).Warn("Skipping event: Cannot build notification for message.")
+		return err
 	}
+
+	batch = append(batch, notification)
+	log.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification in message batch")
+
 	if len(batch) > 0 {
```
tosan88: I think this could be simplified as there will be only one notification after all

ioanaciont: @tosan88 Thanks Sandor, it has been removed.

resources/healthchecks.go
```diff
@@ -1,108 +1,53 @@
...
+		w.Write([]byte("Kafka healthcheck failed"))
 	}
 }
 
-// Check is the the NotificationsPushHealthcheck method that checks is the queue proxies are available
+// Check is the the NotificationsPushHealthcheck method that checks if the kafka queue is available
 func (hc *NotificationsPushHealthcheck) Check() fthealth.Check {
 	return fthealth.Check{
 		BusinessImpact:   "Notifications about newly modified/published content will not reach this app, nor will they reach its clients.",
 		Name:             "MessageQueueProxyReachable",
```
davidbalazs93: should we rename this healthcheck to MessageQueueReachable? :)

app.go
```diff
@@ -41,35 +42,17 @@ func main() {
 		EnvVar: "NOTIFICATIONS_RESOURCE",
 	})
 	consumerAddrs := app.String(cli.StringOpt{
-		Name:   "consumer_proxy_addr",
+		Name:   "consumer_addr",
 		Value:  "",
 		Desc:   "Comma separated kafka proxy hosts for message consuming.",
```
davidbalazs93: remove the "proxy" word from description

PR #48, opened 2017-08-04
Adding Title and Scoop fields to notifications

ioanaciont:
Adding Title and Scoop fields to notifications

### DIFF
--- a/consumer/mapper_test.go
+++ b/consumer/mapper_test.go
@@ -8,8 +8,8 @@ import (
 )
 
 func TestMapToUpdateNotification(t *testing.T) {
-
-	payload := struct{ Foo string }{"bar"}
+	standout := map[string]interface{}{"scoop": true}
+	payload := map[string]interface{}{"title": "This is a title", "standout": standout}
 
 	event := PublicationEvent{
 		ContentURI:   "http://list-transformer-pr-uk-up.svc.ft.com:8081/list/blah/" + uuid.NewV4().String(),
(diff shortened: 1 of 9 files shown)


ioanaciont:
[Add 'title' (headline) and 'scoop' to Push Notifications](https://trello.com/c/56nzCl7X/37-add-title-headline-and-scoop-to-push-notifications)

dispatcher/notifications.go
```diff
@@ -8,6 +8,8 @@ type Notification struct {
 	PublishReference string `json:"publishReference,omitempty"`
 	LastModified     string `json:"lastModified,omitempty"`
 	NotificationDate string `json:"notificationDate,omitempty"`
+	Title            string `json:"title,omitempty"`
+	Scoop            *bool  `json:"scoop,omitempty"`
```
tosan88: I guess this variable is a  pointer to distinct cases when scoop variable is present vs it's false. I'm curios if B2B would like to have that field available regardless if it's populated or not. (a mental note for me)

ioanaciont: That's right, I used the poiter so that `omitempty` could be applied when the field in not populated. Anyway, we can have a chat about this and decide what's the best approach.

consumer/mapper_test.go
```diff
@@ -47,6 +49,8 @@ func TestMapToUpdateNotification_ForContentWithVersion3UUID(t *testing.T) {
 
 	assert.Equal(t, "http://www.ft.com/thing/ThingChangeType/UPDATE", n.Type, "It is an UPDATE notification")
 	assert.Nil(t, err, "The mapping should not return an error")
+	assert.Nil(t, err, "The mapping should not return an error")
```
davidbalazs93: I think this line was added again by mistake, can you remove it? :)

ioanaciont: @davidbalazs93 Done

PR #50, opened 2017-08-23
Filter notifications by content type

ioanaciont:
Clients can filter the content adding a request param for content type (e.g. https://api.ft.com/content/notifications-push?type=article)
For the moment, we do not support multiple types in a single request -> agreed with David L. (`type=article&type=content` does not work as expected, it will filter results based on the first value from the request -> article)

- If the type is not specified, the filter will be set to `type=Article`
- Clients can specify `type=All` that will return all the notifications, despite of their type.
- `DELETE` notifications will be always sent to the clients, because their payload is empty (in Kafka) and we cannot extract the type

Supported types in notifications push are: `Article`, `Content` (for content placeholders), `ContentPackage`.

### DIFF
--- a/consumer/consume.go
+++ b/consumer/consume.go
@@ -57,9 +57,11 @@ func (qHandler *simpleMessageQueueHandler) HandleMessage(queueMsg kafka.FTMessag
 		log.WithField("transaction_id", msg.TransactionID()).WithField("msg", string(msg.Body)).WithError(err).Warn("Skipping event: Cannot build notification for message.")
 		return err
 	}
-
 	log.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification received")
-	qHandler.dispatcher.Send(notification)
+
+	notificationContentType := qHandler.mapper.GetNotificationContentType(pubEvent)
+	inbound := dispatcher.NewNotificationInbound(notification, notificationContentType)
+	qHandler.dispatcher.Send(inbound)
 
 	return nil
 }
--- a/consumer/consume.go
+++ b/consumer/consume.go
@@ -57,11 +57,9 @@ func (qHandler *simpleMessageQueueHandler) HandleMessage(queueMsg kafka.FTMessag
 		log.WithField("transaction_id", msg.TransactionID()).WithField("msg", string(msg.Body)).WithError(err).Warn("Skipping event: Cannot build notification for message.")
 		return err
 	}
-	log.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification received")
 
-	notificationContentType := qHandler.mapper.GetNotificationContentType(pubEvent)
-	inbound := dispatcher.NewNotificationInbound(notification, notificationContentType)
-	qHandler.dispatcher.Send(inbound)
+	log.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification received")
+	qHandler.dispatcher.Send(notification)
 
 	return nil
 }
--- a/consumer/mapper.go
+++ b/consumer/mapper.go
@@ -42,7 +42,6 @@ func (n NotificationMapper) MapNotification(event PublicationEvent, transactionI
 		LastModified:     event.LastModified,
 		Title:            title,
 		Standout:         dispatcher.Standout{Scoop: scoop},
-		ContentType:      n.GetNotificationContentType(event),
 	}, nil
 }
 
--- a/consumer/mapper.go
+++ b/consumer/mapper.go
@@ -26,12 +26,18 @@ func (n NotificationMapper) MapNotification(event PublicationEvent, transactionI
 	var eventType string
 	var scoop bool
 	var title = ""
+	var contentType = ""
 
 	if event.HasEmptyPayload() {
 		eventType = "DELETE"
 	} else {
 		eventType = "UPDATE"
-		title, scoop = extractDataFromPayload(event)
+		notificationPayloadMap, ok := event.Payload.(map[string]interface{})
+		if ok {
+			title = getValueFromPayload("title", notificationPayloadMap)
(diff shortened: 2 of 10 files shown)


resources/push.go
```diff
@@ -11,10 +11,13 @@ import (
...
 
 const (
-	apiKeyHeaderField = "X-Api-Key"
-	apiKeyQueryParam  = "apiKey"
+	apiKeyHeaderField  = "X-Api-Key"
+	apiKeyQueryParam   = "apiKey"
+	defaultContentType = "Article"
 )
 
+var supportedContentTypes = []string{"Article", "Content", "ContentPackage", "All"}
```
davidbalazs93: it would be nice to have this array as a const one, agree? :)

ioanaciont: @davidbalazs93 I was also thinking about this, but then I found that you can't declare constant arrays in Go. The explanation I found is that `Go consts are things that are evaluated at compile-time. Array values don't exist at compile-time.` 
More on this topic: https://stackoverflow.com/questions/13137463/declare-a-constant-array

dispatcher/subscribers.go
```diff
@@ -45,6 +49,17 @@ func (s *standardSubscriber) Since() time.Time {
 	return s.sinceTime
 }
 
+func (s *standardSubscriber) matchesContentType(n Notification, contentType string) bool {
+	if n.Type == "DELETE" || strings.ToLower(s.acceptedContentType) == "all" {
+		return true
+	}
+	if strings.ToLower(s.acceptedContentType) == strings.ToLower(contentType) {
```
kovacshuni: Here it's enough if you return the equality condition, without the `if`, `true` and `false`.

kovacshuni:
I don't know why is not the `contentType` inside the `Notification` struct. We have multiple levels now. I know maybe you wanted to leave the `Notification` struct to remain compatible with the `/notifications` format, but i don't know if it's worth it. You could discuss with others, Next and PO's.

Very nicely written description (of the PR).

dispatcher/subscribers.go
ioanaciont: @kovacshuni Yep you're right, I'm going to change it :)

ioanaciont:
Thanks @kovacshuni! Is there a way to include `contentType` as a field to `Notification` struct, without exposing it in the json response? I've tried it, but I couldn't find anything useful on this topic. 

As far as I know we don't want to send `contentType` to the Push API clients, because we try to keep the messages quick & short. Also, adding this field might require additional processing at client side, and we would like to avoid it if possible.

dispatcher/subscribers.go
```diff
@@ -53,11 +53,8 @@ func (s *standardSubscriber) matchesContentType(n Notification, contentType stri
 	if n.Type == "DELETE" || strings.ToLower(s.acceptedContentType) == "all" {
 		return true
 	}
-	if strings.ToLower(s.acceptedContentType) == strings.ToLower(contentType) {
-		return true
-	}
 
-	return false
+	return (strings.ToLower(s.acceptedContentType) == strings.ToLower(contentType))
```
kovacshuni: no need for the outer `(`,`)`s :)

dispatcher/notifications.go
```diff
@@ -12,6 +12,7 @@ type Notification struct {
 	Standout         Standout `json:"standout"`
 }
 
+// Standout model for a Notification
```
MihaiSorinMoisa: The constant `changeType` from this file is unused and should be removed.

dispatcher/notifications.go
```diff
@@ -12,6 +12,7 @@ type Notification struct {
...
 type Standout struct {
```
MihaiSorinMoisa: The constant `changeType` from this file is unused and should be removed.

resources/push.go
bogdanguranda: Do you know if we're sending notifications for ImageSet, MediaResource, Video, also? If we do, we should add them here also.

consumer/consume.go
```diff
@@ -57,9 +57,11 @@ func (qHandler *simpleMessageQueueHandler) HandleMessage(queueMsg kafka.FTMessag
 		log.WithField("transaction_id", msg.TransactionID()).WithField("msg", string(msg.Body)).WithError(err).Warn("Skipping event: Cannot build notification for message.")
```
MihaiSorinMoisa: This flow should be tested as well.

consumer/mapper.go
```diff
@@ -67,3 +67,17 @@ func extractDataFromPayload(event PublicationEvent) (string, bool) {
...
 	return title, scoop
 }
+
+// GetNotificationContentType extracts the content type for a PublicationEvent
+func (n NotificationMapper) GetNotificationContentType(event PublicationEvent) string {
+	if event.HasEmptyPayload() {
+		return ""
+	}
+
+	notificationPayloadMap, ok := event.Payload.(map[string]interface{})
```
MihaiSorinMoisa: The following 2 flows should be tested as well.

resources/push.go
ioanaciont: @bogdanguranda No, we don't send notifications for those. _Supported types in notifications push are: Article, Content (for content placeholders), ContentPackage._

bogdanguranda: Ok, thank you, good to know :)

ioanaciont:
@davidbalazs93 @bogdanguranda @MihaiSorinMoisa @kovacshuni I moved `contentType` field inside the `Notification` struct, based on Huni's suggestion - I hope the code is a bit cleaner now. I also updated the implementation & the unit tests considering all the other suggestions. Thanks all for being involved, I really appreciate it :)

kovacshuni:
looking good

ioanaciont:
[Avoid notifying push clients about content they can't use](https://trello.com/c/s1Fg6LxJ/38-avoid-notifying-push-clients-about-content-they-cant-use)

PR #51, opened 2017-08-30
Filter notifications by content type - part 2

tosan88:
Supported types are "Article", "ContentPackage"and "All" which includes CPHs as well.
Also returning client error (400) for unsupported types.

### DIFF
--- a/resources/push_test.go
+++ b/resources/push_test.go
@@ -123,6 +123,42 @@ func TestPushFailed(t *testing.T) {
 	assert.Equal(t, http.StatusInternalServerError, w.Code)
 }
 
+func TestPushInvalidType(t *testing.T) {
+	d := new(MockDispatcher)
+	d.On("Register", mock.AnythingOfType("*dispatcher.standardSubscriber")).Return()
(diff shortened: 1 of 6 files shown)


tosan88:
[Avoid notifying push clients about content they can't use](https://trello.com/c/s1Fg6LxJ/38-avoid-notifying-push-clients-about-content-they-cant-use)

resources/push_test.go
```diff
@@ -123,6 +123,42 @@ func TestPushFailed(t *testing.T) {
...
+
+	httpClient := initializeMockHTTPClient(&mockTransport{
+		responseStatusCode: http.StatusOK,
+	})
+	Push(d, "http://dummy.ft.com", httpClient)(w, req)
+
+	assert.Equal(t, http.StatusBadRequest, w.Code)
+
+	reader := bufio.NewReader(w.Body)
+	body, _ := reader.ReadString(byte(rune(0))) // read to EOF
```
davidbalazs93: I'm not quite sure what the meaning of rune(0) is

tosan88: True that, I removed it and elsewhere in the test file

PR #53, opened 2017-10-11
Adjustment for Content placeholders for Wordpress

kovacshuni:
Title and standout will be populated only for types that are not Content, so Article and ContentPackage.

https://trello.com/c/wdCp9dsc/130-content-placeholders-for-wordpress-content


kovacshuni:
After the discussions regarding content placeholders for wordpress, we agreed there are inconsistencies, so i will not bother removing titles or standouts. We will fix the whole idea later in a more structured way.

PR #54, opened 2017-10-27
Remove configuration for list-notifications-push as the service is no…

tamas-molnar:
… longer needed
- to do only after LNP dependency is removed from PAM


tosan88:
Actually, done by #59

PR #56, opened 2018-01-11
Helm improvements (affinity, memory, healthcheck)

ioanaciont:


helm/notifications-push/values.yaml
```diff
@@ -9,6 +9,8 @@ image:
   repository: coco/notifications-push
   pullPolicy: IfNotPresent
 resources:
+  requests:
+    memory: 100Mi
```
peteclark-ft: This requests limit could probably be lower - `notifications-push` seems to run consistently at ~40Mi and spikes to ~100Mi, so I think the requests could be set to around the 50Mi mark? What do you think?

peteclark-ft: Similarly, the `list-notifications-push` limits could be lower

ioanaciont: I had another look in Grafana chart, and I think you are right - I'll change it to 50Mi.

tosan88: Hi @peteclark-ft @ioanaciont ! In fact, what we have found out, that sometimes the grafana board doesn't show the exact metric values unless zoomed in to the peak. We had an OOM in K8s PROD for notifications-push as well so I would like to request to increase the memory both for limit and request ^^

PR #60, opened 2018-02-02
Improvements

tosan88:
Changes include:

* Handling some errors during consuming messages as fatal errors
* Improved logging
* Improved formatting
* Other small improvements

NOTE: 
  * ~~Some tests are failing~~
  * ~~no tests for the new feature~~
  * also depending on underlying lib's PR: https://github.com/Financial-Times/kafka-client-go/pull/14


dispatch/dispatcher.go
```diff
@@ -1,4 +1,4 @@
-package dispatcher
+package dispatch
```
tamas-molnar: Why this change now? Makes the diff very big.

tosan88: Yeah, maybe I should have left it out

tosan88:
:+1: for changes by @tamas-molnar

resources/validator.go
```diff
@@ -7,29 +7,31 @@ import (
...
 
 	req.Header.Set(apiKeyHeaderField, providedApiKey)
 
-	//if the api key has more than four characters we want to log the first four
+	//if the api key has more than 8 characters we want to log the first and last four
 	apiKeyFirstChars := ""
-	if len(providedApiKey) > 4 {
-		apiKeyFirstChars = providedApiKey[0:4]
+	apiKeyLastChars := ""
+	if len(providedApiKey) > 8 {
```
peteclark-ft: What's the value in logging the last 4 digits of the API key?

resources/validator.go
```diff
@@ -7,29 +7,31 @@ import (
...
 	if err != nil {
-		log.WithField("url", masheryApiKeyValidationURL).WithError(err).Error("Invalid URL for api key validation")
+		log.WithField("url", apiGatewayApiKeyValidationURL).WithError(err).Error("Invalid URL for api key validation")
 		return false, "Invalid URL", http.StatusInternalServerError
 	}
 
 	req.Header.Set(apiKeyHeaderField, providedApiKey)
 
-	//if the api key has more than four characters we want to log the first four
+	//if the api key has more than 8 characters we want to log the first and last four
```
peteclark-ft: What is the standard length for tyk API keys? I'm just not sure this is overly secure.. What if there are 5 or 9 chars in the api key? An attacker with this log line would then have a ~1/62 chance of correctly guessing the remaining character.

Should this be based on a percentage of the length of the key, capped at 4 chars?

supervisor_test.go
```diff
@@ -0,0 +1,74 @@
...
+			atomic.AddInt32(&handlerCalls, 1)
+		}
+
+		supervisor := newServiceSupervisor(serviceName, errCh, fatalErrs, fatalErrHandler)
+		go supervisor.Supervise()
+		for _, err := range tc.inputErrs {
+			errCh <- err
+		}
+
+		time.Sleep(250 * time.Millisecond)
```
peteclark-ft: V Minor, but you could probably speed up this test by using a WaitGroup in the handler or something similar

resources/validator.go
```diff
@@ -43,10 +45,10 @@ func isValidApiKey(providedApiKey string, masheryApiKeyValidationURL string, htt
...
 
 	if respStatusCode == http.StatusUnauthorized {
-		log.WithField("apiKeyFirstChars", apiKeyFirstChars).Error("Invalid api key")
+		log.WithField("apiKeyFirstChars", apiKeyFirstChars).WithField("apiKeyLastChars", apiKeyLastChars).Error("Invalid api key")
 		return false, "Invalid api key", http.StatusUnauthorized
 	}
 
-	log.WithField("url", req.URL.String()).Errorf("Received unexpected status code from Mashery: %d", respStatusCode)
+	log.WithField("url", req.URL.String()).WithField("apiKeyFirstChars", apiKeyFirstChars).WithField("apiKeyLastChars", apiKeyLastChars).Errorf("Received unexpected status code from Api Gateway: %d", respStatusCode)
 	return false, "Request to validate api key returned an unexpected response", http.StatusServiceUnavailable
```
peteclark-ft: Should we not also be forwarding `429` and `403` errors to the client here? As we know, these are legitimate responses from Tyk for overuse of API keys.

resources/validator.go
tamas-molnar: The gateway team recommend we log the last 5 chars of a key, since the new Tyk keys have identical prefixes. This will be fixed here https://github.com/Financial-Times/notifications-push/compare/key-validation-fixes?expand=1

resources/validator.go
tamas-molnar: Yes we should, that will be fixed in https://github.com/Financial-Times/notifications-push/compare/key-validation-fixes?expand=1

PR #73, opened 2018-10-22
Filter DELETE messages by content type header

ioanaciont:


consumer/consume.go
```diff
@@ -83,6 +83,7 @@ func (qHandler *simpleMessageQueueHandler) HandleMessage(queueMsg kafka.FTMessag
 		}
 	}
 
+	pubEvent.ContentTypeHeader = contentType
```
bulicamihai: Here could be a problem because of contentType. ContentType could be for example "application/vnd.ft-upp-audio+json; version=1.0; charset=utf-8", if you take a look on master, it is a method that is striping contentType until ";".

PR #89, opened 2020-03-02
Fix/UPPSF-1108 Address growing number of code issues

ivan-p-nikolov:
List of  addressed changes to notifications-push:

- [Clean Dispatcher interface](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r374079038)
- [Rename Register interface](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r374080607)
- [Change Dispatcher Send method to receive a single notification](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r374085313)
- [Consolidate route registration code](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r374169703)
- [Cleanup mock definitions](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r374224244)
- [Gracefully shutdown http server](https://github.com/Financial-Times/notifications-push/pull/85#discussion_r376067030)
- Move to the new logger library


dispatch/dispatcher_test.go
```diff
@@ -233,7 +260,7 @@ func TestInternalFailToSendNotifications(t *testing.T) {
 	d.Send(n1)
 
 	time.Sleep(time.Second)
-	logger.Info("This log message is here to avoid a race condition")
+	//logger.Info("This log message is here to avoid a race condition")
```
MiroslavGatsanoga: Let's remove the code instead of commenting it out.

mocks/mocks.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return nil
+}
+
+// Dispatcher is a mock of a dispatcher that can be reused for testing
+type Dispatcher struct {
+	mock.Mock
+}
+
+// Start mocks Start
```
MiroslavGatsanoga: Let's remove such doc comments (there are a couple more below) since they are not adding any value to the code.

resources/healthchecks_test.go
```diff
@@ -60,134 +36,58 @@ func TestHealthcheck(t *testing.T) {
...
-				},
-			},
-			expectedStatus: 200,
-			expectedBody:   "pong",
-		},
 	}
 
 	for name, test := range tests {
-		fmt.Printf("Running test %s \n", name)
+		t.Run(name, func(t *testing.T) {
```
MiroslavGatsanoga: In order to run all test cases capture the range variable i.e. `test := test`.

resources/http_client.go
```diff
@@ -2,26 +2,25 @@ package resources
...
 import (
 	"bytes"
-	"errors"
+	"fmt"
 	"net/http"
 	"time"
 )
 
-type HttpClient struct {
+type HTTPClient struct {
```
MiroslavGatsanoga: The type has no internal state, so we can remove it and use a function instead of the method below.

consumer/mapper.go
```diff
@@ -25,6 +25,7 @@ var UUIDRegexp = regexp.MustCompile("[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4
 func (n NotificationMapper) MapNotification(event PublicationEvent, transactionID string) (dispatch.Notification, error) {
 	UUID := UUIDRegexp.FindString(event.ContentURI)
 	if UUID == "" {
+		// nolint:golint
```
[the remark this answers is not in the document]
ivan-p-nikolov: I found it a good idea to add the `nolint` to show that it was intentionally left like this.

app_integration_test.go
```diff
@@ -63,8 +62,8 @@ var annotationMsg = kafka.NewFTMessage(map[string]string{
 
 func TestPushNotifications(t *testing.T) {
 
-	l := logger.NewUPPLogger("TEST", "PANIC")
-	log.InitLogger("TEST", "PANIC")
+	l := logger.NewUPPLogger("TEST", "info")
```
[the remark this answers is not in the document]
ivan-p-nikolov: If the log level is at `panic`, the logger would not create the log entry objects for lesser logs. So log hooks and such would not fire. 
My purpose for setting the log level to `panic` was to stop the logger from polluting test logs. Discarding the output achieves the same goal, and setting the log level to `info` makes the test more representative of real world setup.

consumer/content_test.go
```diff
@@ -30,8 +30,7 @@ func TestSyntheticMessage(t *testing.T) {
 	msg := kafka.NewFTMessage(map[string]string{"X-Request-Id": "SYNTH_tid"},
 		`{"UUID": "a uuid", "ContentURI": "http://list-transformer-pr-uk-up.svc.ft.com:8080/lists/blah/55e40823-6804-4264-ac2f-b29e11bf756a"}`)
 
-	err := handler.HandleMessage(msg)
-	assert.NoError(t, err)
+	assert.NoError(t, handler.HandleMessage(msg))
```
epavlova, line 33: I would have kept the call to the function that is being tested on a separate line from the error assertion as it's easier for me to read the test that way.

consumer/content_test.go
```diff
@@ -86,18 +96,104 @@ func TestSparkCCTWhitelist(t *testing.T) {
...
+		"fail to map event to notification": {
+			Headers:     map[string]string{"X-Request-Id": "tid_test", "Content-Type": "application/json"},
+			Body:        `{"UUID": "a uuid", "ContentURI": "http://upp-article-mapper.svc.ft.com:8080/content/uuid"}`,
+			ExpectError: true,
+			Message:     "Skipping event: Cannot build notification for message.",
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
```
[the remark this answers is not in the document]
MiroslavGatsanoga: The capturing of the range variable should happen in the `for` loop body, not in the anonymous function i.e. 
```
for name, test := range tests {
		test := test
		t.Run(name, func(t *testing.T) {
                ...
```

consumer/metadata_test.go
```diff
@@ -87,13 +88,15 @@ func TestMetadata(t *testing.T) {
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
```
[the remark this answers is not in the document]
MiroslavGatsanoga: The capturing of the range variable should happen in the `for` loop body, not in the anonymous function i.e. 
```
for name, test := range tests {
		test := test
		t.Run(name, func(t *testing.T) {
                ...
```

resources/healthchecks_test.go
MiroslavGatsanoga: The capturing of the range variable should happen in the `for` loop body, not in the anonymous function i.e. 
```
for name, test := range tests {
		test := test
		t.Run(name, func(t *testing.T) {
                ...
```

PR #90, opened 2020-04-14
Feature/uppsf 1303 change message format

ivan-p-nikolov:


consumer/mapper.go
```diff
@@ -59,15 +54,24 @@ func (n NotificationMapper) MapNotification(event PublicationEvent, transactionI
...
+	if event.Payload == nil {
+		return dispatch.Notification{}, errors.New("payload missing")
+	}
+
 	return dispatch.Notification{
 		Type:             dispatch.AnnotationUpdateType,
-		ID:               "http://www.ft.com/thing/" + event.ContentID,
-		APIURL:           n.APIBaseURL + "/" + n.Resource + "/" + event.ContentID,
+		ID:               "http://www.ft.com/thing/" + UUID,
+		APIURL:           n.APIBaseURL + "/" + n.Resource + "/" + UUID,
```
epavlova, line 70: [I don't think that this should be part of this PR, I just noticed it.] As `n.APIBaseURL` is something that is given to the service as input when it is started, it could have `/` or not and we can't guarantee that. Using concatenation for constructing url path may end up with having "//" in the string result. We could overcome this by constructing the url here with a bit more complex solution:
- parse the `APIBaseURL` as `url.URL` object
- construct `URL.Path` using `path.Join`
- parse the `url.URL` as string again

consumer/publish_event.go
```diff
@@ -35,20 +35,26 @@ func (msg NotificationQueueMessage) OriginID() string {
...
-	var event ConceptAnnotationsEvent
+// AsMetadata converts message to ConceptAnnotations Event
+func (msg NotificationQueueMessage) AsMetadata() (AnnotationsMessage, error) {
+	var event AnnotationsMessage
 	err := json.Unmarshal([]byte(msg.Body), &event)
 	return event, err
 }
 
-type ConceptAnnotationsEvent struct {
+type AnnotationsMessage struct {
```
karsov, line 51: I don't think that the new name is a good idea.
The old name reflects better the fact that this represents a "PostConceptAnnotations" message.

consumer/publish_event.go
```diff
@@ -58,23 +64,22 @@ type ConceptAnnotationsEvent struct {
 	} `json:"annotations"`
 }
 
-// PublicationEvent is the data structure that represents a publication event consumed from Kafka
-type PublicationEvent struct {
+// ContentMessage is the data structure that represents a publication event consumed from Kafka
+type ContentMessage struct {
```
karsov, line 68: I don't think that the new name is a good idea.
The old name reflects better the fact that this represents a "CmsPublicationEvents" message.

consumer/publish_event.go
```diff
@@ -35,20 +35,26 @@ func (msg NotificationQueueMessage) OriginID() string {
...
 }
 
-type ConceptAnnotationsEvent struct {
+type AnnotationsMessage struct {
+	ContentURI   string       `json:"contentUri"`
+	LastModified string       `json:"lastModified"`
+	Payload      *Annotations `json:"payload"`
+}
+
+type Annotations struct {
```
karsov, line 57: This is used only in a test. I don't see value in it not being simply an empty interface.

epavlova:
It was easy for me to review and understand the code with the new names of the structs. The separate types and the separate mappings in the mappings.go is clean for me. It's just opinion, I don't have strong arguments.

PR #97, opened 2020-11-30
Enable e2e capability tests

MiroslavGatsanoga:
# Description

Implement sending E2E test notifications to monitoring subscribers.

## What

Sending the E2E test notifications to monitoring subscribers is done according to a configurable allow list of UUIDs.
The checks that PAM schedules regarding the Push notifications capability now produce metrics indicating success i.e. status metric with value 1 or time metric lower than 120 (our SLA).
The metrics can be found at (https://graphitev2-api.ft.com/) under `Metrics/content-metadata/capability-monitoring directory`.

## Why

[JIRA ticket](https://financialtimes.atlassian.net/browse/UPPSF-1917). 

- [ ] Bug fix
- [X] Feature
- [ ] Documentation
- [ ] Breaking change

___


ivan-p-nikolov (dismissed):
Looks good to me. I would have tried to extract the content UUID from the message body instead of the transaction ID, but maybe that's more work for not a lot of gain.

Can we document in the README `Monitoring` section the new functionality? The fact that we have UUIDs that would not generate notifications for normal subscribers, only monitor ones. Also that the UUID is extracted from the transaction and not the body.

consumer/publish_event.go
```diff
@@ -19,6 +19,18 @@ func (msg NotificationQueueMessage) HasSynthTransactionID() bool {
 	return strings.HasPrefix(tid, "SYNTH")
 }
 
+// HasE2ETestTransactionID checks if the message is e2e test
+func (msg NotificationQueueMessage) HasE2ETestTransactionID(e2eTestUUIDs []string) bool {
+	tid := msg.TransactionID()
+	for _, testUUID := range e2eTestUUIDs {
+		if strings.Contains(tid, testUUID) {
```
ivan-p-nikolov, line 26: I was wondering. Should we handle lower/upper case letters for the UUIDs? Currently only metadata messages have upper case letters, but still.

ivan-p-nikolov (approved):

consumer/publish_event.go
MiroslavGatsanoga: I don't see the need for that now but it can be done in the future.

mitagg (approved):

MiroslavGatsanoga:
Rebased onto master, I think it's ok to disregard the Snyk check fail as I suspect it is an issue with Snyk.

mitagg (approved):
We can deal with Snyk later, LGTM!

ivan-p-nikolov (approved):

PR #107, opened 2021-05-18
Add support for delete notification from payload flag

gotha:


consumer/mapper.go
```diff
@@ -29,7 +29,7 @@ func (n NotificationMapper) MapNotification(event ContentMessage, transactionID
 	var title = ""
 	var contentType = ""
 
-	if event.HasEmptyPayload() {
+	if event.HasEmptyPayload() || event.HasDeleteFlag() {
```
atanasdinov, line 32: `event.HasEmptyPayload()` call (and the function itself) are now obsolete with the new event structure.

consumer/mapper_test.go
```diff
@@ -77,6 +77,28 @@ func TestMapToDeleteNotification(t *testing.T) {
...
+	}
+
+	mapper := NotificationMapper{
+		APIBaseURL: "test.api.ft.com",
+		Resource:   "list",
+	}
+
+	n, err := mapper.MapNotification(event, "tid_test1")
+
+	assert.Equal(t, "http://www.ft.com/thing/ThingChangeType/DELETE", n.Type, "It is an UPDATE notification")
```
atanasdinov, line 98: The describing message should be "It is a DELETE notification"

consumer/publish_event.go
```diff
@@ -90,6 +92,16 @@ func (e ContentMessage) Matches(whitelist *regexp.Regexp) bool {
 	return whitelist.MatchString(e.ContentURI)
 }
 
+// HasDeleteFlag - checks if payload has delete flag
+func (e ContentMessage) HasDeleteFlag() bool {
+	data, ok := e.Payload.(map[string]interface{})
```
atanasdinov, line 97: This works however I dislike the hidden code duplication that it introduces.

We have the following piece of code within the `MapNotification` function:

```
if event.HasEmptyPayload() || event.HasDeleteFlag() {
	eventType = dispatch.ContentDeleteType
	contentType = resolveTypeFromMessageHeader(event.ContentTypeHeader)
} else {
	eventType = dispatch.ContentUpdateType
	notificationPayloadMap, ok := event.Payload.(map[string]interface{})
	if ok {
		title = getValueFromPayload("title", notificationPayloadMap)
		contentType = getValueFromPayload("type", notificationPayloadMap)
		scoop = getScoopFromPayload(notificationPayloadMap)
	}
}
```
 
Each received message results in a call to `HasDeleteFlag()` which would do the type assertion we see here. However, if we aren't dealing with delete event we will do the same assertion again as seen in the quoted snippet.

The better approach here would be to extract the type assertion prior to checking for the event type and if it fails, we should return an error. You can follow the changes [here](https://github.com/Financial-Times/content-exporter/pull/32/files) for a similar example.

atanasdinov (changes requested):
The changes are working however they can be simplified.

Also please adjust (or remove altogether) the existing test case(s) for delete messages with empty payload - that structure is no longer supported by our services.

atanasdinov (dismissed):

gotha:
the second commit message has nothing to do with the actual change. 
Why is the snyk scan removed ?

gotha:
actually the last commit message is only partially related to tests and I would say it is far from the most important change

atanasdinov (approved):

AlexKondov (approved):

PR #111, opened 2021-06-22
implement create event notification based on subscriber policy

Zarrie:
https://financialtimes.atlassian.net/browse/UPPSF-2473

In order to test the changes you must subscribe for notifications using the following command 

```curl -i --header "x-api-key:<key>" --header "X-API-Policy: INTERNAL_UNSTABLE" https://api-t.ft.com/content/notifications-push?type=All```

and create new content via the cms notifier **(on staging)** including the following property

```
{
    "_internal": {
        "publishCount": 1
    },
...
}
```
this will emit CREATE notitification.

### DIFF
--- a/consumer/content.go
+++ b/consumer/content.go
@@ -96,7 +96,11 @@ func (qHandler *ContentQueueHandler) HandleMessage(queueMsg kafka.FTMessage) err
 	}
 	notification.IsE2ETest = isE2ETest
 
-	monitoringLogger.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification received")
+	qHandler.log.
+		WithField("resource", notification.APIURL).
+		WithField("transaction_id", notification.PublishReference).
+		WithField("notification_type", notification.Type).
+		Info("Valid notification received")
 	qHandler.dispatcher.Send(notification)
 
 	return nil
--- a/consumer/content.go
+++ b/consumer/content.go
@@ -96,9 +96,8 @@ func (qHandler *ContentQueueHandler) HandleMessage(queueMsg kafka.FTMessage) err
 	}
 	notification.IsE2ETest = isE2ETest
 
-	qHandler.log.
+	monitoringLogger.
 		WithField("resource", notification.APIURL).
-		WithField("transaction_id", notification.PublishReference).
 		WithField("notification_type", notification.Type).
 		Info("Valid notification received")
 	qHandler.dispatcher.Send(notification)
--- a/consumer/content_test.go
+++ b/consumer/content_test.go
@@ -80,7 +80,7 @@ func TestSparkCCTWhitelist(t *testing.T) {
 	}
 	l := logger.NewUPPLogger("test", "PANIC")
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, NewSet(), nil, mapper, dispatcher, l)
 
@@ -104,7 +104,7 @@ func TestMonitoringEvents(t *testing.T) {
 	defer h.Reset()
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	typeWhitelist := NewSet()
 	typeWhitelist.Add("valid-type")
@@ -183,7 +183,7 @@ func TestAcceptNotificationBasedOnContentType(t *testing.T) {
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 	l := logger.NewUPPLogger("test", "PANIC")
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(defaultContentURIWhitelist, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -207,7 +207,7 @@ func TestAcceptNotificationBasedOnAudioContentType(t *testing.T) {
 	contentTypeWhitelist.Add("application/vnd.ft-upp-audio+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(defaultContentURIWhitelist, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -231,7 +231,7 @@ func TestDiscardNotificationBasedOnContentType(t *testing.T) {
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -256,7 +256,7 @@ func TestAcceptNotificationBasedOnContentUriWhenContentTypeIsApplicationJson(t *
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -280,7 +280,7 @@ func TestDiscardNotificationBasedOnContentUriWhenContentTypeIsApplicationJson(t 
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -305,7 +305,7 @@ func TestAcceptNotificationBasedOnContentUriWhenContentTypeIsMissing(t *testing.
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -328,7 +328,7 @@ func TestDiscardNotificationBasedOnContentUriWhenContentTypeIsMissing(t *testing
 	contentTypeWhitelist.Add("application/vnd.ft-upp-article+json")
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(sparkIncludedWhiteList, contentTypeWhitelist, nil, mapper, dispatcher, l)
 
@@ -369,7 +369,7 @@ func TestHandleMessage(t *testing.T) {
 	}
 	l := logger.NewUPPLogger("test", "PANIC")
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(defaultContentURIWhitelist, NewSet(), nil, mapper, dispatcher, l)
 
@@ -460,7 +460,7 @@ func TestAcceptNotificationBasedOnE2ETransactionID(t *testing.T) {
 	e2eTestUUIDs := []string{"e4d2885f-1140-400b-9407-921e1c7378cd"}
 	l := logger.NewUPPLogger("test", "PANIC")
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationModel")).Return()
 
 	handler := NewContentQueueHandler(nil, nil, e2eTestUUIDs, mapper, dispatcher, l)
 
--- a/consumer/mapper.go
+++ b/consumer/mapper.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strconv"
 
 	"github.com/Financial-Times/notifications-push/v5/dispatch"
 )
@@ -17,7 +18,10 @@ type NotificationMapper struct {
 // UUIDRegexp enables to check if a string matches a UUID
 var UUIDRegexp = regexp.MustCompile("[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}")
 
-const payloadDeletedFlag = "deleted"
+const (
+	payloadDeletedFlag = "deleted"
+	publishCountKey    = "publishCount"
+)
 
 // MapNotification maps the given event to a new notification.
 func (n NotificationMapper) MapNotification(event ContentMessage, transactionID string) (dispatch.Notification, error) {
@@ -43,7 +47,23 @@ func (n NotificationMapper) MapNotification(event ContentMessage, transactionID 
 		eventType = dispatch.ContentDeleteType
 		contentType = resolveTypeFromMessageHeader(event.ContentTypeHeader)
 	} else {
-		eventType = dispatch.ContentUpdateType
+		publishCount, ok := notificationPayloadMap[publishCountKey]
+		var publishCountValue int
+		if ok {
+			publishCount := fmt.Sprintf("%v", publishCount)
+			var err error
+			publishCountValue, err = strconv.Atoi(publishCount)
+			if err != nil {
+				return dispatch.Notification{}, fmt.Errorf("invalid value for field publishCount in: %T", publishCount)
+			}
+		}
+
+		if publishCountValue == 1 {
+			eventType = dispatch.ContentCreateType
+		} else {
+			eventType = dispatch.ContentUpdateType
+		}
+
 		title = getValueFromPayload("title", notificationPayloadMap)
 		contentType = getValueFromPayload("type", notificationPayloadMap)
 		scoop = getScoopFromPayload(notificationPayloadMap)
--- a/consumer/mapper.go
+++ b/consumer/mapper.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-	"strconv"
 
 	"github.com/Financial-Times/notifications-push/v5/dispatch"
 )
@@ -19,16 +18,16 @@ type NotificationMapper struct {
 var UUIDRegexp = regexp.MustCompile("[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}")
 
 const (
-	payloadDeletedFlag = "deleted"
-	publishCountKey    = "publishCount"
+	payloadDeletedKey      = "deleted"
+	payloadPublishCountKey = "publishCount"
 )
 
 // MapNotification maps the given event to a new notification.
-func (n NotificationMapper) MapNotification(event ContentMessage, transactionID string) (dispatch.Notification, error) {
+func (n NotificationMapper) MapNotification(event ContentMessage, transactionID string) (dispatch.NotificationModel, error) {
 	UUID := UUIDRegexp.FindString(event.ContentURI)
 	if UUID == "" {
 		// nolint:golint
-		return dispatch.Notification{}, errors.New("ContentURI does not contain a UUID")
+		return dispatch.NotificationModel{}, errors.New("ContentURI does not contain a UUID")
 	}
 
 	var eventType string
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: consumer/content.go [hunk 1]
```diff
@@ -96,7 +96,11 @@ func (qHandler *ContentQueueHandler) HandleMessage(queueMsg kafka.FTMessage) err
 	}
 	notification.IsE2ETest = isE2ETest
 
-	monitoringLogger.WithField("resource", notification.APIURL).WithField("transaction_id", notification.PublishReference).Info("Valid notification received")
+	qHandler.log.
```
gotha, line 99: why using `qHandler.log` instead of `monitoringLogger` ?

consumer/mapper.go
```diff
@@ -43,7 +47,23 @@ func (n NotificationMapper) MapNotification(event ContentMessage, transactionID
 		eventType = dispatch.ContentDeleteType
 		contentType = resolveTypeFromMessageHeader(event.ContentTypeHeader)
 	} else {
-		eventType = dispatch.ContentUpdateType
+		publishCount, ok := notificationPayloadMap[publishCountKey]
+		var publishCountValue int
```
gotha, line 51: you can simplify the code block like this 
```
	eventType = dispatch.ContentUpdateType
	publishCountStr, ok := notificationPayloadMap[publishCountKey].(string)
	if ok {
		publishCount, err := strconv.Atoi(publishCountStr)
		if err != nil {
			return dispatch.Notification{}, fmt.Errorf("could not cast publishCount value to int")
		}

		if publishCount == 1 {
			eventType = dispatch.ContentCreateType
		}

     }
```


or even simpler like this:
```
eventType = dispatch.ContentUpdateType
publishCountStr, _ := notificationPayloadMap[publishCountKey].(string)
if publishCountStr == "1" {
	eventType = dispatch.ContentCreateType
}
```

dispatch/notifications.go
```diff
@@ -2,19 +2,21 @@ package dispatch
...
-	AllContentType         = "All"
+	AnnotationsType           = "Annotations"
+	ArticleContentType        = "Article"
+	ContentPackageType        = "ContentPackage"
+	AudioContentType          = "Audio"
+	LiveBlogPackageType       = "LiveBlogPackage"
+	LiveBlogPostType          = "LiveBlogPost"
+	ContentPlaceholderType    = "Content"
+	AllContentType            = "All"
+	CreateEventConsideredType = "INTERNAL_UNSTABLE"
```
gotha, line 13: `INTERNAL_UNSTABLE` is not subscription type.
If anything it should be considered subscription configuration separate from the subscription type. 
Also, why is it defined in `dispatch` package and is only used on `resources` package ?

dispatch/subscribers.go
```diff
@@ -30,27 +31,29 @@ type NotificationConsumer interface {
...
-	notificationChannel chan string
-	addr                string
-	sinceTime           time.Time
-	acceptedTypes       []string
+	id                      string
+	notificationChannel     chan string
+	addr                    string
+	sinceTime               time.Time
+	acceptedTypes           []string
+	sendCreateNotifications bool
```
gotha, line 39: I think it is better to abstract this a little bit better this. 
If we another notification type (like majorUpdate for example) I dont want to have to have to add another flag called `sendMajorUpdateNotifications` or something like that

dispatch/notifications.go
```diff
@@ -38,3 +40,24 @@ type Notification struct {
 type Standout struct {
 	Scoop bool `json:"scoop"`
 }
+
+func consolidateNotificationType(notification Notification, isCreateAllowed bool) Notification {
```
gotha, line 44: the name of the function does not describe accurately what the function does.
It creates new notification. 
Maybe instead of creating new Notification we can use different `views` of a single struct.

gotha (changes requested):

consumer/content.go
Zarrie: mistake

consumer/mapper.go
Zarrie: I didn't want to do this, because I consider bad practice not treating values with their intended types as you suggest, but if you are okay with that and even suggesting it I'm fine doing it - it's not a strong preference anyway

dispatch/subscribers.go
Zarrie: I would strongly disagree with this as it's the very definition of premature optimisation. Just as in this PR if we have to add something we do the proper refactoring to enable the change. In this case it would be just replacing the variable with slice of policies, but I see no reason to do the refactoring now and add complexity that may never be needed.

consumer/mapper.go
```diff
@@ -17,7 +18,10 @@ type NotificationMapper struct {
 // UUIDRegexp enables to check if a string matches a UUID
 var UUIDRegexp = regexp.MustCompile("[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}")
 
-const payloadDeletedFlag = "deleted"
+const (
+	payloadDeletedFlag = "deleted"
+	publishCountKey    = "publishCount"
```
atanasdinov, line 23: I'd like to suggest a more unified way of defining the name for these constants. Both of those are used to check the payload of a message yet one is prefixed with `payload` whereas the other one is not, one is called `Flag` whereas the other one is `Key`.

Perhaps add a comment above the block to indicate the following constants are related to the mssage payload and keep the `nameKey` pattern so that `payloadDeletedFlag` is renamed to `deletedKey` or whatever you prefer.

resources/push.go
```diff
@@ -18,27 +18,29 @@ const (
...
-	strings.ToLower(dispatch.ContentPlaceholderType): true,
+	strings.ToLower(dispatch.AnnotationsType):           true,
+	strings.ToLower(dispatch.ArticleContentType):        true,
+	strings.ToLower(dispatch.ContentPackageType):        true,
+	strings.ToLower(dispatch.AudioContentType):          true,
+	strings.ToLower(dispatch.AllContentType):            true,
+	strings.ToLower(dispatch.LiveBlogPackageType):       true,
+	strings.ToLower(dispatch.LiveBlogPostType):          true,
+	strings.ToLower(dispatch.ContentPlaceholderType):    true,
+	strings.ToLower(dispatch.CreateEventConsideredType): true,
```
atanasdinov, line 35: Only supported content types are listed here. Configuration for `CREATE` events shouldn't be included.

resources/push.go
```diff
@@ -182,10 +189,15 @@ func getClientAddr(r *http.Request) string {
...
+
 	values := r.URL.Query()
 	subTypes := values["type"]
 	if len(subTypes) == 0 {
-		return []string{defaultSubscriptionType}, nil
+		if len(retVal) == 0 {
+			return []string{defaultSubscriptionType}, nil
+		}
+		retVal = append(retVal, defaultSubscriptionType)
+		return retVal, nil
```
atanasdinov, line 200: The behaviour is the same but the logic is just more complex. What is the reasoning behind the change?

dispatch/subscribers.go
atanasdinov: Perhaps we could just extract an `eventSbscriptionConfig` and use the options "pattern" in order to fill it. Thoughts?

resources/push_test.go
```diff
@@ -34,34 +34,40 @@ func TestSubscription(t *testing.T) {
 		ExpectedBody   string
 		ExpectedStatus int
 		ExpectStream   bool
+		Policies       map[dispatch.SubscriberPolicy]bool
+		RequestHeader  string
```
atanasdinov, line 38: I believe this naming is too generic and further updates to the tests might be confusing. Do you think `SubscriptionPolicy` is better?

resources/push_test.go
```diff
@@ -148,9 +177,9 @@ func TestPassKeyAsParameter(t *testing.T) {
 	v := &mocks.KeyValidator{}
 	v.On("Validate", mock.Anything, keyAPI).Return(nil)
 
-	sub, _ := dispatch.NewStandardSubscriber(req.RemoteAddr, []string{defaultSubscriptionType})
+	sub, _ := dispatch.NewStandardSubscriber(req.RemoteAddr, []string{defaultSubscriptionType}, map[dispatch.SubscriberPolicy]bool{dispatch.CreateEventPolicy: false})
```
atanasdinov, line 180: Is there a benefit of all these map definitions here and below? Each function could define a single map and use it on each function call:

`policies := map[dispatch.SubscriberPolicy]bool{}`

These tests have no association with the subscription policies and there's also no difference between setting the policy to `false` and omitting it altogether.

dispatch/notifications.go
```diff
@@ -34,7 +35,43 @@ type Notification struct {
...
+	APIURL           string    `json:"apiUrl"`
+	ID               string    `json:"id"`
+	Type             string    `json:"type"`
+	SubscriberID     string    `json:"subscriberId,omitempty"`
+	PublishReference string    `json:"publishReference,omitempty"`
+	LastModified     string    `json:"lastModified,omitempty"`
+	NotificationDate string    `json:"notificationDate,omitempty"`
+	Title            string    `json:"title,omitempty"`
+	Standout         *Standout `json:"standout,omitempty"`
+	SubscriptionType string    `json:"-"`
```
atanasdinov, line 49: This and the `IsE2ETest` fields are not needed because they are not included with the marshalling.

app_integration_test.go
```diff
@@ -293,8 +293,8 @@ func testClientWithNONotifications(ctx context.Context, t *testing.T, serverURL
 				return
 			case body := <-ch:
 				delta := time.Since(start)
-				assert.InEpsilon(t, heartbeat.Nanoseconds(), delta.Nanoseconds(), 0.05, "No Notification Client with type '%s' expects to receive heatbests on time.")
-				assert.Equal(t, "data: []\n\n", body, "No Notification Client with type '%s' expects to receive only heartbeat messages.")
+				assert.InEpsilon(t, heartbeat.Nanoseconds(), delta.Nanoseconds(), 0.05, "No NotificationModel Client with type '%s' expects to receive heatbests on time.")
+				assert.Equal(t, "data: []\n\n", body, "No NotificationModel Client with type '%s' expects to receive only heartbeat messages.")
```
atanasdinov, line 297: Are these changes intended?

consumer/content_test.go
```diff
@@ -104,7 +104,7 @@ func TestMonitoringEvents(t *testing.T) {
 	defer h.Reset()
 
 	dispatcher := &mocks.Dispatcher{}
-	dispatcher.On("Send", mock.AnythingOfType("dispatch.Notification")).Return()
+	dispatcher.On("Send", mock.AnythingOfType("dispatch.NotificationResponse")).Return()
```
atanasdinov, line 107: Could you please verify these tests are actually working? The mock defines the following function:

```
func (m *Dispatcher) Send(notification dispatch.NotificationModel) {
	m.Called(notification)
}
```

consumer/mapper.go
atanasdinov: Is this about to be changed according to the suggestions above or will it be left as is?

dispatch/dispatcher.go
```diff
@@ -145,24 +145,24 @@ func (d *Dispatcher) forwardToSubscribers(notification Notification) {
...
-			WithField("resource", notification.APIURL)
+			WithTransactionID(nm.PublishReference).
+			WithField("resource", nm.APIURL)
 
-		if notification.IsE2ETest {
+		if nm.IsE2ETest {
 			if _, isStandard := sub.(*StandardSubscriber); isStandard {
 				skipped++
-				entry.Info("Test notification. Skipping standard subscriber.")
+				entry.Info("Test nm. Skipping standard subscriber.")
```
atanasdinov, line 154: Was this log message automatically changed?

dispatch/dispatcher_test.go
```diff
@@ -66,8 +85,8 @@ func TestShouldDispatchNotificationsToMultipleSubscribers(t *testing.T) {
 	h := NewHistory(historySize)
 	d := NewDispatcher(delay, h, l)
 
-	m, _ := d.Subscribe("192.168.1.2", contentSubscribeTypes, true)
-	s, _ := d.Subscribe("192.168.1.3", contentSubscribeTypes, false)
+	m, _ := d.Subscribe("192.168.1.2", contentSubscribeTypes, true, map[SubscriberPolicy]bool{CreateEventPolicy: false})
```
atanasdinov, line 88: These tests have no association with the subscription policies and there's also no difference between setting the policy to `false` and omitting it altogether.

It should be possible to use the same map (since it's read-only) for all subscribers as it simplifies the subscription calls: 

`policies := map[dispatch.SubscriberPolicy]bool{}`

(Doesn't apply to `TestCreateNotificationIsProperlyDispatchedToSubscribers`.)

consumer/mapper_test.go
```diff
@@ -34,6 +34,32 @@ func TestMapToUpdateNotification(t *testing.T) {
 	assert.Equal(t, "Article", n.SubscriptionType, "SubscriptionType field should be mapped correctly")
 }
 
+func TestMapToCreateNotification(t *testing.T) {
+	t.Parallel()
+
+	standout := map[string]interface{}{"scoop": true}
+	payload := map[string]interface{}{"title": "This is a title", "standout": standout, "type": "Article", "publishCount": 1}
```
atanasdinov, line 41: We could benefit from one extra test case where the `publishCount` is not 1 to verify we'd receive an `UPDATE` notification.

atanasdinov (changes requested):

consumer/mapper.go
Zarrie: Well, I personally find it more easy to aim for simple and descriptive name, though I do not state those are, instead of consistent across some scope names. Also simple clicking on the variables shows where they are used as they are used on a single place, so I don't actually think this introduces any complexity. Also their use is restricted to the given file only.

resources/push.go
```diff
@@ -15,11 +15,13 @@ import (
...
-	apiKeyQueryParam        = "apiKey"
-	ClientAdrKey            = "X-Forwarded-For"
-	defaultSubscriptionType = dispatch.ArticleContentType
+	HeartbeatMsg                  = "[]"
+	apiKeyHeaderField             = "X-Api-Key"
+	apiKeyQueryParam              = "apiKey"
+	apiPolicyField                = "X-API-Policy"
+	ClientAdrKey                  = "X-Forwarded-For"
+	defaultSubscriptionType       = dispatch.ArticleContentType
+	createEventSubscriptionHeader = "INTERNAL_UNSTABLE"
```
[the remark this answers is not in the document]
Zarrie: as the follow up task has to introduce validation of the policies and so on this change will actually be removed and properly validated if the required policies are present. This is more like a mock for testing and does not matter at all since the policies are going to be retrieved via apig soon.

dispatch/subscribers.go
```diff
@@ -95,26 +109,41 @@ func (s *StandardSubscriber) Send(n Notification) error {
...
+		return nil, err
+	}
+
+	return buffer.Bytes(), err
+}
+
+// MarshalNotificationModelsJSON returns the JSON encoding of n. For notification models, we do not use the standard function json.Marshal()
 // because that will always escape special characters (<,>,&) in unicode format ("\u0026P" and similar)
-func MarshalNotificationsJSON(n []Notification) ([]byte, error) {
+func MarshalNotificationModelsJSON(n []NotificationModel) ([]byte, error) {
```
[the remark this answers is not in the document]
Zarrie: Do you suggest changing the model with the response structure in the history? If not I'm not sure what you suggest. If yes it does not quite make sense because the response is representation depending on the subscriber while I would expect the history to store the actual notification either create or update agnostic of what the subscriber has subscribed for

resources/push_test.go
Zarrie: I'm generally not a fan of premature optimisation as this is the very definition of it handling cases which might actually never occur, because confusion would arise only if something else called policies or similar would be introduced, but as this would cost nothing I will change it :)

Zarrie:
@atanasdinov please tag the team after a review so they have the chance to discuss and leave comments as well

consumer/mapper.go
gotha: I would prefer to name it either `payloadPublishCountKey`. 
It may be better if `payloadDeletedFlag` is renamed to `payloadDeletedKey` for consistency, but I personally do not think it is that important.

dispatch/dispatcher.go
```diff
@@ -74,13 +74,13 @@ func (d *Dispatcher) Subscribers() []Subscriber {
 	return subs
 }
 
-func (d *Dispatcher) Subscribe(address string, subTypes []string, monitoring bool) (Subscriber, error) {
+func (d *Dispatcher) Subscribe(address string, subTypes []string, monitoring bool, policies map[SubscriberPolicy]bool) (Subscriber, error) {
```
gotha, line 77: I would suggest to use `[]SubscrberPolicy` instead of `map[SubscriberPolicy]bool`.
If you want to mimic Set in go you should use `map[type]struct{}`

gotha: `SubscriberPolicy` terminology is confusing because it sounds like it is related to API policy. 
For me it would me cleared if we use something like `SubscriptionOption`

dispatch/notifications.go
```diff
@@ -15,12 +15,13 @@ const (
...
 const (
 	ContentUpdateType    = "http://www.ft.com/thing/ThingChangeType/UPDATE"
+	ContentCreateType    = "http://www.ft.com/thing/ThingChangeType/CREATE"
 	ContentDeleteType    = "http://www.ft.com/thing/ThingChangeType/DELETE"
 	AnnotationUpdateType = "http://www.ft.com/thing/ThingChangeType/ANNOTATIONS_UPDATE"
 )
 
-// Notification model
-type Notification struct {
+// NotificationModel model
```
gotha, line 23: I personally would not rename this to `NotificationModel`, but if you insist to name it that way to distinguish it from `NotificationsResponse` I will squint really hard

dispatch/notifications.go
```diff
@@ -15,12 +15,13 @@ const (
...
+type NotificationModel struct {
 	APIURL           string    `json:"apiUrl"`
```
gotha, line 25: Is there a  need to keep the `json` tags ?

dispatch/notifications.go
```diff
@@ -34,7 +35,43 @@ type Notification struct {
...
+	SubscriptionType string    `json:"-"`
+	IsE2ETest        bool      `json:"-"`
+}
+
+// Standout model for a NotificationResponse
 type Standout struct {
 	Scoop bool `json:"scoop"`
 }
+
+func createNotificationResponse(notification NotificationModel, subscriber Subscriber) NotificationResponse {
```
gotha, line 58: I would suggest instead of using the whole subscriber to check if it has specific policy, just pass the policies (a.k.a options) instead. This way it would be slightly more decoupled

app_integration_test.go
```diff
@@ -293,7 +293,7 @@ func testClientWithNONotifications(ctx context.Context, t *testing.T, serverURL
 				return
 			case body := <-ch:
 				delta := time.Since(start)
-				assert.InEpsilon(t, heartbeat.Nanoseconds(), delta.Nanoseconds(), 0.05, "No Notification Client with type '%s' expects to receive heatbests on time.")
+				assert.InEpsilon(t, heartbeat.Nanoseconds(), delta.Nanoseconds(), 0.05, "No NotificationModel Client with type '%s' expects to receive heatbests on time.")
```
gotha, line 296: does this message need to be changed ?

dispatch/dispatcher.go
```diff
@@ -145,24 +145,24 @@ func (d *Dispatcher) forwardToSubscribers(notification Notification) {
...
-			if !matchesSubType(notification, sub) {
+			if !matchesSubType(nm, sub) {
 				skipped++
 				entry.Info("Skipping subscriber.")
 				continue
 			}
 		}
-
-		err := sub.Send(notification)
+		nr := createNotificationResponse(nm, sub)
```
gotha, line 164: ideally pass policies instead of the whole subscriber .

dispatch/subscribers.go
```diff
@@ -14,18 +14,25 @@ const notificationBuffer = 16
 
 var ErrSubLagging = errors.New("subscriber lagging behind")
 
+type SubscriberPolicy string
+
+const (
+	CreateEventPolicy SubscriberPolicy = "CreateEventPolicy"
```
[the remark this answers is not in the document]
gotha: I would personally type it like this 
```
	CreateEventPolicy SubscriberPolicy = "CreateEvent"
```

resources/history.go
```diff
@@ -13,7 +13,7 @@ func History(history dispatch.History, log *logger.UPPLogger) func(w http.Respon
 	return func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-type", "application/json; charset=UTF-8")
 
-		historyJSON, err := dispatch.MarshalNotificationsJSON(history.Notifications())
+		historyJSON, err := dispatch.MarshalNotificationModelsJSON(history.Notifications())
```
gotha, line 16: I would expect `/__history` to convert `NotificationModel` to `NotificationReponse`. 
At this point we would not be accepting specific policies and we will return always `UPDATE` but in the future when we want to add this feature it would be trivial.

dispatch/subscribers.go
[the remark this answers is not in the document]
gotha: :+1:
please check my comment on `resources/history.go`

resources/push.go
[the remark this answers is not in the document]
gotha: I am fine with how it is right now.

gotha (changes requested):

resources/push_test.go
```diff
@@ -28,48 +28,74 @@ func TestSubscription(t *testing.T) {
...
-		ExpectedStatus int
-		ExpectStream   bool
+		Request            string
+		IsMonitor          bool
+		ExpectedType       []string
+		ExpectedBody       string
+		ExpectedStatus     int
+		ExpectStream       bool
+		SubscriptionPolicy map[dispatch.SubscriberPolicy]bool
+		RequestHeader      string
```
gotha, line 38: I would personally use either `map[string]string` to denote headers or `map[string][]string` to match http.Request Header definition

dispatch/dispatcher.go
Zarrie: hasOption is method of the Subscriber interface. I'd prefer to keep it that way instead of enabling passing only the policies but changing the logic.

dispatch/notifications.go
Zarrie: "hasOption is method of the Subscriber interface. I'd prefer to keep it that way instead of enabling passing only the policies but changing the logic."

If you really insist and consider it better quality code I will do it anyway

consumer/mapper.go
```diff
@@ -34,22 +37,27 @@ func (n NotificationMapper) MapNotification(event ContentMessage, transactionID
...
 
-	deleteFlag, _ := notificationPayloadMap[payloadDeletedFlag].(bool)
+	deleteFlag, _ := notificationPayloadMap[payloadDeletedKey].(bool)
 
 	if deleteFlag {
 		eventType = dispatch.ContentDeleteType
 		contentType = resolveTypeFromMessageHeader(event.ContentTypeHeader)
 	} else {
 		eventType = dispatch.ContentUpdateType
+		publishCountStr := fmt.Sprintf("%v", notificationPayloadMap[payloadPublishCountKey])
```
gotha, line 50: imo, using sprintf to cast to string is (at least ugly). 
I would suggest to use `notificationPayloadMap[payloadPublishCountKey].(string)` as it was originally suggested

dispatch/history.go
```diff
@@ -36,14 +36,18 @@ func (i *inMemoryHistory) Push(n Notification) {
 	}
 }
 
-func (i *inMemoryHistory) Notifications() []Notification {
+func (i *inMemoryHistory) Notifications() []NotificationResponse {
```
gotha, line 39: storage and presentation logic are mixed here imo. 

inMemoryHistory should accept,store and return `NotificationModel`. 
History should not care about how this data is represented. 
The conversion from `NotificationModel` to `NotificationResponse` should happen in the `history` resource

dispatch/notifications.go
```diff
@@ -31,11 +48,55 @@ type Notification struct {
...
+		SubscriberID:     notification.SubscriberID,
+		PublishReference: notification.PublishReference,
+		LastModified:     notification.LastModified,
+		NotificationDate: notification.NotificationDate,
+		Title:            notification.Title,
+		Standout:         notification.Standout,
+	}
+}
+
+func notificationModelToResponse(nm NotificationModel) NotificationResponse {
```
gotha, line 86: there is no need for this function, just use `createNotificationResponse` instead. 
There is no point to duplicate the same logic twice

dispatch/notifications.go
```diff
@@ -31,11 +48,55 @@ type Notification struct {
...
+		if o == option {
+			return true
+		}
+	}
+	return false
+}
+
+func createNotificationResponse(notification NotificationModel, subscriberOptions []SubscriptionOption) NotificationResponse {
+	notificationType := notification.Type
+	if notificationType == ContentCreateType && !hasOption(CreateEventOption, subscriberOptions) {
```
gotha, line 69: the function (and the name)`hasOption` implies there is some specific business logic in finding `option`. 
The is not, what is implemented is `array.exists(item)`, go std lib does not have such function on purpose. 
The go way is to use `for`. 
I would personally remove `hasOption` function

atanasdinov (dismissed):

dispatch/notifications.go
```diff
@@ -31,11 +48,36 @@ type Notification struct {
...
 	Scoop bool `json:"scoop"`
 }
+
+func CreateNotificationResponse(notification NotificationModel, subscriberOptions []SubscriptionOption) NotificationResponse {
+	notificationType := notification.Type
+
+	hasCreateEventOption := false
+	for _, o := range subscriberOptions {
+		if o == CreateEventOption {
+			hasCreateEventOption = true
```
gotha, line 64: it is a good practice to break after you have found what you are looking for

resources/history.go
```diff
@@ -13,7 +13,12 @@ func History(history dispatch.History, log *logger.UPPLogger) func(w http.Respon
 	return func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-type", "application/json; charset=UTF-8")
 
-		historyJSON, err := dispatch.MarshalNotificationsJSON(history.Notifications())
+		var notifications []dispatch.NotificationResponse
+		for _, n := range history.Notifications() {
+			notifications = append(notifications, dispatch.CreateNotificationResponse(n, []dispatch.SubscriptionOption{}))
```
gotha, line 18: you can just use `nil` as second parameter to `CreateNotificationResponse`

dispatch/notifications.go
gotha: this is not resolved. 
Why is `CreateEventConsideredType` is defined in `dispatch` package and used only in `resources` package

atanasdinov (approved):

gotha (approved):

gotha:
LGTM, lets do some more manual testing and I think we are ready to go to prod

PR #114, opened 2021-07-22
Subscribe for advanced notifications based on API key policies

atanasdinov:
# Description

## What

Use API key policies in order to subscribe for advanced notifications.

## Why

[JIRA Ticket](https://financialtimes.atlassian.net/browse/UPPSF-2507)

* Perhaps `KeyProcessor` isn't the best name to use for the new functionality so I'm open to suggestions.
* The `validator.go` and respectively `validator_test.go` modules will be renamed to `key_processor(_test).go` (or something else if suggested) before merging but are left like this to allow for easier review of the changes - otherwise Git recognises those as brand new files.
* The commits will also be rebased before merging.

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Breaking change

___


resources/validator.go
```diff
@@ -63,55 +77,119 @@ func (v *KeyValidator) Validate(ctx context.Context, key string) error {
...
+
+	req, err := http.NewRequestWithContext(ctx, "GET", p.policiesURL, nil)
+	if err != nil {
+		return nil, NewKeyErrWithDescription("Invalid policies URL", http.StatusInternalServerError, "", err.Error())
+	}
+
+	req.Header.Set(APIKeyHeaderField, key)
+
+	// Use the last characters of the key for logging purposes.
+	var keySuffix string
```
StephanZahariev, line 147: What if we use:
```
keySuffix := key
```

This will log something for the key even if the length is bellow the specified suffix length

atanasdinov: I assume we'd never have such a short key (less than 10 characters) however potentially exposing a key with its full length to the logs might be problematic if later down the road these keys can also have "write" permissions. Thanks for bringing to light.

@gotha what do you think about this suggestion?

StephanZahariev (dismissed):

resources/validator.go
```diff
@@ -63,55 +77,119 @@ func (v *KeyValidator) Validate(ctx context.Context, key string) error {
...
+			errMsg = "API key policies not found"
+		}
+
+		return nil, NewKeyErrWithDescription(errMsg, resp.StatusCode, keySuffix, string(data))
+	}
+
+	// YAML parsing is used because the response of the API
+	// is a "relaxed" JSON (e.g. { 'x-policy': ''}) which
+	// cannot be parsed by the standard JSON unmarshaler.
+	keyPolicies := struct {
```
gotha, line 177: parsing "json" as yaml is just weird. 
If it cannot be parsed properly, I would suggest to use regex
```
	r := regexp.MustCompile(`x-policy':\s+'(.*)?'`)
	res := r.FindStringSubmatch("{ 'x-policy': 'orel1,orel2'}")
```

Zarrie (approved):

gotha (changes requested):

gotha (approved):

gotha (approved):

Zarrie (approved):

PR #133, opened 2022-11-07
Feature/uppsf 3271 extract annotation notifications

martin-stanchev:
# Description

## What

Separating the annotations notifications logic and simplifying the service to be more configurable.

## Why

https://financialtimes.atlassian.net/browse/UPPSF-3271

## Anything, in particular, you'd like to highlight to reviewers

Annotations messages were previously validated against the origin, but since they contain `contentURI`s I've removed that logic and added a content uri allowlist for annotations coming from pac.annotations-rw-neo4j and next-video.annotations-neo4j. Let me know if this creates any problems.

- [x] Tech hygiene (dependency updating & other tech debt)
- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Breaking change

___


dispatch/dispatcher.go
```diff
@@ -154,7 +154,7 @@ func (d *Dispatcher) forwardToSubscribers(notification NotificationModel) {
 				entry.Info("Test notification. Skipping standard subscriber.")
 				continue
 			}
-		} else {
+		} else if notification.SubscriptionType != "" {
```
martin-stanchev, line 157: The `subscriptionType` variable is set from the Content Type and since it's not present in annotation kafka messages, we should skip the validation. This also will be skipped for lists and pages, where only 1 type is validated and returned.

consumer/consume.go
```diff
@@ -14,30 +17,106 @@ type notificationDispatcher interface {
...
+	if h.monitorsEvents {
+		logEntry = h.log.WithMonitoringEvent("NotificationsPush", tid, pubEvent.ContentType)
+	} else {
+		logEntry = h.log.WithTransactionID(tid)
+	}
+
+	if err != nil {
+		logEntry.WithError(err).Error("Failed to unmarshall kafka message")
 		return
 	}
```
ilian2233, line 79: **Question:** Is it possible to have a monitoring event that fails to unmarshal? I think the code will panic in that case.

If true I think the issue can be solved by something like:
```suggestion
        logEntry := h.log.WithTransactionID(tid)
	pubEvent, err := msg.Unmarshal()
	if err != nil {
		logEntry.WithError(err).Error("Failed to unmarshall kafka message")
		return
	}
	
	if h.monitorsEvents {
		logEntry = h.log.WithMonitoringEvent("NotificationsPush", tid, pubEvent.ContentType)
        }	
```

martin-stanchev: I initially had the log initialising the same way that you suggested, but there are [these test cases](https://github.com/Financial-Times/notifications-push/blob/feature/UPPSF-3271-extract-annotation-notifications/consumer/consumer_test.go#L144) which expect a monitoring event log when the kafka message body is not able to be unmarshalled. If you look at the [old implementation](https://github.com/Financial-Times/notifications-push/blob/master/consumer/content.go#L57), it's also initialised in the same way.

I think this is because if some messages are failing to be processed the monitoring will not be able to catch them if there are no monitoring events and in this case the `pubEvent.ContentType` is set from the kafka header, so in theory there shouldn't be any issues.

consumer/consume.go
```diff
@@ -14,30 +17,106 @@ type notificationDispatcher interface {
...
 }
 
-type MessageQueueRouter struct {
-	contentHandler  MessageQueueHandler
-	metadataHandler MessageQueueHandler
+var exists = struct{}{}
+
+type Set struct {
+	m map[string]struct{}
+}
```
ilian2233, line 24: **nitpick (non-blocking):** This can be replaced with `type Set map[string]struct{}`

ilian2233 (approved):
LGTM!

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
```diff
@@ -0,0 +1,20 @@
+# Values used for the deployed application.
+replicaCount: 2
+service:
+  name: annotations-notifications-push
```
dtvalk-ov, line 4: Wondering if we should be consistent with the other deployments and use singular form for the service name (annotation-notifications-push) like list-* and page-* 🤔 
For me it sounds better as annotationS-*, but for the sake of consistency.. 😄

martin-stanchev: I don't mind either. I named it plural because other annotations services like `annotations-rw-neo4j`, `annotations-writer-ontotext` are named plural. I guess it's a case of what we want to be consistent with 😄

consumer/content.go
```diff
@@ -1,113 +0,0 @@
...
-
-	if !isE2ETest && notification.SubscriptionType == dispatch.ArticleContentType {
-		qHandler.log.WithField("eventType", notification.Type).WithField("ID", notification.ID).WithTransactionID(tid).Info("Processed article notification")
-	}
-	qHandler.dispatcher.Send(notification)
-}
-
-func stripDirectives(contentType string) string {
-	return strings.Split(contentType, ";")[0]
-}
```
ivanruski, line 113: **question:** Was that used to handle some old contentType ?

consumer/consume.go
```diff
@@ -14,30 +17,106 @@ type notificationDispatcher interface {
...
+	}
+
+	isE2ETest := msg.HasE2ETestTransactionID(h.e2eTestUUIDs)
+	if !isE2ETest {
+		if msg.HasSynthTransactionID() {
+			logEntry.WithValidFlag(false).WithField("contentUri", pubEvent.ContentURI).Info("Skipping event: Synthetic transaction ID.")
+			return
+		}
+
+		if pubEvent.ContentType == "application/json" {
```
ivanruski, line 93: **question:** We will enter this `if` only when the message type is annotations and the `else` for the other instances of the service, right ?

ivanruski (approved):
LGTM.

>Annotations messages were previously validated against the origin, but since they contain contentURIs I've removed that logic and added a content uri allowlist for annotations coming from pac.annotations-rw-neo4j and next-video.annotations-neo4j. Let me know if this creates any problems.

This should not create any problems. `annotations-rw-neo4j` will set the content uri to `http://pac.annotations-rw-neo4j.svc.ft.com/annotations/uuid` when the origin is http://cmdb.ft.com/systems/pac and `http://next-video.annotations-rw-neo4j.svc.ft.com/annotations/uuid` when the origin is http://cmdb.ft.com/systems/next-video-editor. For any other kind of origins the message won't be ingested.

dispatch/dispatcher.go
ivanruski: **suggestion (non-blocking):** If we add `MessageType` field to `NotificationMessage` struct, we could use it in `mapper.go` to distinguish the subscription type. In this way we won't treat empty `dispatch.NotificationModel.SubscriptionType` as `dispatch.AnnotationsType`. We would have to check both `Content-Type` and `Message-Type` headers to determine the subscription type though.

consumer/consume.go
martin-stanchev: Yes, in the previous implementation there was also an empty string check, which i think should not be possible in theory. 

Another thought I had was that video (as the content, not the annotations) also is saved as application/json for now, but there are no notifications for videos currently. I guess if that changes this should still work fine.

consumer/content.go
martin-stanchev: As far as i can tell there were some content types (coming from methode i think) that were in the format `application/vnd.ft-upp-article+json; utf8 ...` so this was to remove those 'directives'. 

I haven't found any messages with content types like that so I don't think we need it.

helm/notifications-push/app-configs/notifications-push_eks_delivery.yaml
```diff
@@ -4,15 +4,18 @@ service:
   name: notifications-push
 env:
   TOPIC: "PublicationNotificationEvents"
-  METADATA_TOPIC: "PostConceptAnnotations"
   NOTIFICATIONS_RESOURCE: "content"
-  CONTENT_URI_WHITELIST: "^http://(methode|wordpress-article|content|upp)(-collection|-content-placeholder|-notifications-creator)?(-mapper|-unfolder|)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content|complementarycontent)/[\\w-]+.*$"
-  CONTENT_TYPE_WHITELIST: "application/vnd.ft-upp-article-internal+json,application/vnd.ft-upp-content-package+json,application/vnd.ft-upp-audio+json,application/vnd.ft-upp-live-blog-post-internal+json,application/vnd.ft-upp-live-blog-package-internal+json"
-  WHITELISTED_METADATA_ORIGIN_SYSTEM_HEADERS: "http://cmdb.ft.com/systems/pac, http://cmdb.ft.com/systems/methode-web-pub, http://cmdb.ft.com/systems/next-video-editor"
+  API_URL_RESOURCE: "content"
+  CONTENT_URI_ALLOWLIST: "^http://(content|upp)(-collection|-content-placeholder|-notifications-creator)?(-mapper|-unfolder|)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content|complementarycontent)/[\\w-]+.*$"
```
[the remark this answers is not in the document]
martin-stanchev: I don't think so. I remember we looked did some research on this when we were rewriting the ingesters, but I don't think we came up with a list of all the URIs. I will check the ingesters though, they are most likely to be up to date

README.md
```diff
@@ -19,60 +19,69 @@ go install
...
-    && export KAFKA_ADDRS=localhost:2181 \
+    && export API_URL_RESOURCE=content \
+    && export KAFKA_ADDRESS=localhost:9092 \
     && export GROUP_ID=notifications-push-yourtest \
     && export TOPIC=PostPublicationEvents \
     && export NOTIFICATIONS_DELAY=10 \
     && export API_BASE_URL="http://api.ft.com" \
-    && export CONTENT_TYPE_WHITELIST="application/vnd.ft-upp-article+json,application/vnd.ft-upp-content-package+json" \
-    && export CONTENT_URI_WHITELIST="^http://(methode|wordpress|content)-(article|collection|content-placeholder)-(transformer|mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content)/[\\w-]+.*$" \
+    && export CONTENT_TYPE_ALLOWLIST="application/vnd.ft-upp-article+json,application/vnd.ft-upp-content-package+json" \
```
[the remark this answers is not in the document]
martin-stanchev: Hmm i guess we don't. They're validated through the article validator with the article schema. I am not sure if it's still present in old published content, but i think it's not needed here anyways

helm/notifications-push/app-configs/notifications-push_eks_delivery.yaml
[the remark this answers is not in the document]
martin-stanchev: I don't think so. I remember we looked did some research on this when we were rewriting the ingesters, but I don't think we came up with a list of all the URIs. I will check the ingesters though, they are most likely to be up to date

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
```diff
@@ -0,0 +1,20 @@
...
+env:
+  TOPIC: "PostConceptAnnotations"
+  NOTIFICATIONS_RESOURCE: "annotations"
+  API_URL_RESOURCE: "content"
+
+  MONITOR_NOTIFICATIONS: "false"
```
atanasdinov, line 10: Are there any monitoring logs for annotations? If such exist we might not need to do this.

consumer/mapper.go
```diff
@@ -22,76 +25,49 @@ const (
...
-		publishCountStr := fmt.Sprintf("%v", notificationPayloadMap[payloadPublishCountKey])
-		if publishCountStr == "1" {
-			eventType = dispatch.ContentCreateType
-		}
-
-		title = getValueFromPayload("title", notificationPayloadMap)
-		contentType = getValueFromPayload("type", notificationPayloadMap)
-		scoop = getScoopFromPayload(notificationPayloadMap)
+	// If it's a create event.
+	if publishCount, ok := notificationPayloadMap[payloadPublishCountKey]; ok && publishCount == "1" {
```
atanasdinov, line 45: This will not work. I guess because the `publishCount` is an integer.

consumer/mapper.go
```diff
@@ -22,76 +25,49 @@ const (
...
+		eventType = dispatch.ContentCreateType
 	}
 
-	var standout *dispatch.Standout
-	if contentType != dispatch.ListType && contentType != dispatch.PageType {
-		standout = &dispatch.Standout{Scoop: scoop}
+	// If it's a delete event.
+	if _, ok := notificationPayloadMap[payloadDeletedKey].(bool); ok {
```
atanasdinov, line 50: I'd prefer to have the previous check or a combination of both instead of only verifying whether such field exists and is boolean. This might cause issues if we want to play around with the `deleted` key in the future.

consumer/mapper.go
```diff
@@ -9,8 +9,11 @@ import (
 
 // NotificationMapper maps CmsPublicationEvents to Notifications
 type NotificationMapper struct {
-	APIBaseURL string
-	Resource   string
+	APIBaseURL      string
+	Resource        string
```
atanasdinov, line 13: This is not used anymore.

README.md
```diff
@@ -283,13 +291,13 @@ How to Build & Run with Docker
     docker build -t coco/notifications-push .
 
     docker run --env NOTIFICATIONS_RESOURCE=content \
-        --env KAFKA_ADDRS=localhost:2181 \
+        --env KAFKA_ADDRESS=localhost:2181 \
```
atanasdinov, line 294:
```suggestion
        --env KAFKA_ADDRESS=localhost:9092 \
```

2181 is the host of the Zookeeper so this hasn't been updated with the MSK migration.

README.md
```diff
@@ -283,13 +291,13 @@ How to Build & Run with Docker
...
         --env GROUP_ID="notifications-push-yourtest" \
         --env TOPIC="PostPublicationEvents" \
         --env NOTIFICATIONS_DELAY=10 \
         --env API_BASE_URL="http://api.ft.com" \
-        --env CONTENT_TYPE_WHITELIST="application/vnd.ft-upp-article+json,application/vnd.ft-upp-content-package+json" \
-        --env CONTENT_URI_WHITELIST="^http://(methode|wordpress|content)-(article|collection)-(transformer|mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content)/[\\w-]+.*$" \
+        --env CONTENT_TYPE_ALLOWLIST="application/vnd.ft-upp-article+json,application/vnd.ft-upp-content-package+json" \
+        --env CONTENT_URI_ALLOWLIST="^http://(methode|wordpress|content)-(article|collection)-(transformer|mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content)/[\\w-]+.*$" \
```
atanasdinov, line 300: Could we please have the updated allowlist from the app config?

helm/notifications-push/app-configs/notifications-push_eks_delivery.yaml
atanasdinov, line 9:
```suggestion
  CONTENT_URI_ALLOWLIST: "^http://(content|upp)(-collection|-content-placeholder|-notifications-creator)?(-mapper|-unfolder|)\\.svc\\.ft\\.com(:\\d{2,5})?/(content|complementarycontent)/[\\w-]+.*$"
```

Can be simplified even further but I'd not take the chances with that many changes in a single PR.

atanasdinov (commented):
Looks great overall! Some things I noticed:
- docker-compose config should also be updated with the `ALLOWLIST` change
- synthetic annotation publications will pass through e.g. `427f2a19-2ae7-47c1-b580-c225fa0a0199`

consumer/consume.go
atanasdinov: There's something fishy about it. We allow notifications for video annotations but not videos themselves. Probably something to address with the BAs?

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
atanasdinov: I'm with the 🐺 on this one. In the other examples we're talking about `annotations` and in this service we're having `annotation notifications` in mind. See `annotation-notifications-ingester`. It's a nitpick tho.

dispatch/dispatcher.go
atanasdinov: The function called in this `else` block i.e. `matchesSubType()` has annotation logic related to it. In any case, a comment in the code would be great to have if we're keeping this check.

helm/notifications-push/app-configs/notifications-push_eks_delivery.yaml
atanasdinov: At a second look, this would always expect a `(-mapper|-unfolder|)`. Is it even working properly prior to this PR?

martin-stanchev: The `|)` probably is what allows that group to be empty 😄 . I can change it to `(-mapper|-unfolder)?` so it's cleaner, it shouldn't affect how it works.

I was also debating about `(-pr|-iw)?(-uk-.*)?` but I'm not sure if it will affect something. As Vankata mentioned it might be worth to collect all the URIs that we can publish content with so we can update these URIs

dispatch/dispatcher.go
ivanruski: Yes, a comment would be fine, because this info will get buried after this PR is merged.

helm/notifications-push/app-configs/notifications-push_eks_delivery.yaml
atanasdinov: Ah, you are probably right about the `|)`... 😄 

`(-pr|-iw)?(-uk-.*)?` was part of the content types published by Methode in the part.

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
martin-stanchev: Do you mean if there's anything checking for monitoring logs for annotations? PAM doesn't perform any specific checks for the annotation notifications and annotations in general. I am not sure if there's anything that will check the monitoring logs, but annotations didn't log with monitoring events before so that's why I added this.

consumer/consume.go
martin-stanchev: I'll ask Annie about it!

dispatch/dispatcher.go
martin-stanchev: I added the suggestion with the `Message-Type` from the kafka header. Check out how that works and tell me if you like it. I think it's better this way. 

The only issue is with the logic of handling delete notifications. From what I understood there are no delete notifications for annotations, but we should allow delete notifications if the subscription type is empty. This is strange and I am not sure if it makes sense even.

dispatch/dispatcher.go
```diff
@@ -181,15 +181,8 @@ func matchesSubType(n NotificationModel, s Subscriber) bool {
...
-	ann := strings.ToLower(AnnotationsType)
 
-	if n.Type == ContentDeleteType {
-		if notifType != "" {
-			return subTypes[notifType]
-		}
-		if len(subTypes) <= 1 && subTypes[ann] {
-			return false
-		}
+	if n.Type == ContentDeleteType && notifType == "" {
```
martin-stanchev, line 185: So `n.Type == ContentDeleteType` should not happen for annotations anyways, and an empty subscription type is also pretty much not possible as far as i can tell

consumer/consume.go
martin-stanchev: So Penko and Annie suggest that we limit video annotation notifications to a x-policy and think about adding video notifications with the same policy in the future. This will make this task much more complex as we don't allow the selection between annotations for different content types, but rather serve all annotations through the endpoint. 

I will add a task that we will discuss on our refinement, but this will go out of scope for this task.

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
atanasdinov: Not if there's something checking for such logs automatically but to run manually a Splunk query for annotation publishes with `monitoring_event=true`. If there are such logs we should be able to change the logging here as well without any negative impacts.

martin-stanchev: Ah okay. Well there seem to be these services -> 

pac-annotations-mapper
annotations-rw-neo4j  
annotations-monitoring-service

The last monitors if the annotations make it to neo4j if i understood it correctly, but I'm not sure if the change here will cause any other issues. @ivanruski what do you think about this?

consumer/mapper.go
```diff
@@ -19,79 +21,57 @@ var UUIDRegexp = regexp.MustCompile("[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4
...
+		eventType = dispatch.ContentDeleteType
+		contentType = resolveTypeFromMessageHeader(event.ContentType)
 	}
 
-	var standout *dispatch.Standout
-	if contentType != dispatch.ListType && contentType != dispatch.PageType {
-		standout = &dispatch.Standout{Scoop: scoop}
+	if contentType == "" && event.MessageType == annotationMessageType {
+		contentType = dispatch.AnnotationsType
 	}
```
ivanruski, line 57: **nitpick (non-blocking):** I think that this is better suited right after the line:
```
contentType := getValueFromPayload("type", notificationPayloadMap)
```

dispatch/dispatcher.go
ivanruski: Can we have `notifType == ""` after the latest change to set `SubscriptionType` based on the Message-Type header?

ivanruski (changes requested):
LGTM! Just one chore for mapper_test.go

martin-stanchev:
@ivanruski I can't reply to your comment for some reason 😄 

> Can we have notifType == "" after the latest change to set SubscriptionType based on the Message-Type header?

Theoretically yes, if Message-Type is not `concept-annotations` and there is no `type` in the kafka body it will be empty. In practice though i don't think this will happen.

README.md
```diff
@@ -78,14 +90,14 @@ export NOTIFICATIONS_RESOURCE=pages \
...
-    --consumer_addr="localhost:2181" \
+    --consumer_addr="localhost:9092" \
     --consumer_group_id="notifications-push" \
     --topic="PostPublicationEvents" \
     --notifications_delay=10 \
     --api_base_url="http://api.ft.com" \
     --api_key_validation_endpoint="t800/a" \
-    --content_type_whitelist="application/vnd.ft-upp-article+json" --content_type_whitelist="application/vnd.ft-upp-content-package+json" \
-    --content_uri_whitelist="^http://(methode|wordpress|content)-(article|collection|content-placeholder)-(transformer|mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content)/[\\w-]+.*$" \
+    --content_type_allowlist="application/vnd.ft-upp-article+json" --content_type_whitelist="application/vnd.ft-upp-content-package+json" \
```
atanasdinov, line 99:
```suggestion
    --content_type_allowlist="application/vnd.ft-upp-article+json" \
```

helm/notifications-push/app-configs/annotation-notifications-push_eks_delivery.yaml
```diff
@@ -0,0 +1,20 @@
...
+  API_URL_RESOURCE: "content"
+
+  MONITOR_NOTIFICATIONS: "false"
+  UPDATE_EVENT_TYPE: "http://www.ft.com/thing/ThingChangeType/ANNOTATIONS_UPDATE"
+
+  CONTENT_URI_ALLOWLIST: "^http://(pac|next-video)\\.annotations-rw-neo4j\\.svc\\.ft\\.com/annotations/[\\w-]+.*$"
+  CONTENT_TYPE_ALLOWLIST: "application/json"
+
+  API_KEY_VALIDATION_ENDPOINT: "t800/a"
+  API_KEY_POLICIES_ENDPOINT: "t800/policy"
```
atanasdinov, line 17: Is this one necessary?

docker-compose.yaml
```diff
@@ -29,14 +29,13 @@ services:
       API_BASE_URL: "http://api-t.ft.com"
       GROUP_ID: "notifications-push-test1"
       TOPIC: "PostPublicationEvents"
-      METADATA_TOPIC: "PostConceptAnnotations"
       NOTIFICATIONS_DELAY: "10"
       NOTIFICATIONS_RESOURCE: "content"
-      CONTENT_URI_WHITELIST: "^http://(methode|wordpress-article|content)(-collection|-content-placeholder)?-(mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content|complementarycontent)/[\\w-]+.*$$"
-      CONTENT_TYPE_WHITELIST: "application/vnd.ft-upp-article+json,application/vnd.ft-upp-content-package+json,application/vnd.ft-upp-audio+json,application/vnd.ft-upp-live-blog-post+json,application/vnd.ft-upp-live-blog-package+json"
+      API_URL_RESOURCE: "content"
+      CONTENT_URI_ALLOWLIST: "^http://(methode|wordpress-article|content)(-collection|-content-placeholder)?-(mapper|unfolder)(-pr|-iw)?(-uk-.*)?\\.svc\\.ft\\.com(:\\d{2,5})?/(content|complementarycontent)/[\\w-]+.*$$"
```
atanasdinov, line 35: Could we also update this allowlist?

ivanruski (dismissed):

ilian2233 (approved):

helm/notifications-push/app-configs/annotations-notifications-push_eks_delivery.yaml
ivanruski: The monitor for annotations publishes cares about 3 kinds of `monitoring_event`s , "PublishStart"(logged in `cms-metadata-notifier`), "Map"(logged in `pac/video-annotations-mapper`), and "SaveNeo4j"(logged in `annotations-rw-neo4j`). I think that this change won't affect the annotations monitoring process.

ivanruski (approved):

atanasdinov (approved):

ilian2233 (approved):

PR #142, opened 2023-09-08
Added central banking filter

ilian2233:
# Description

## What

- Added integration with OPA library
- Added opa module file containing rule for filtering out `central banking` content

## Why

[JIRA ticket](https://financialtimes.atlassian.net/browse/UPPSF-4588)

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Breaking change

___


antoanFT (approved):

app.go
```diff
@@ -196,6 +189,14 @@ func main() {
 			log.WithError(err).Fatalf("could not create Kafka consumer for %s and topic %s", *consumerAddress, *kafkaTopic)
 		}
 
+		evaluator, err := access.CreateEvaluator(
+			"data.centralBanking.allow",
```
martin-stanchev, line 193: **question**: Will there be any benefit in exporting this to the helm configs or are we aiming to extract this later on?

dispatch/dispatcher_test.go
```diff
@@ -87,7 +87,12 @@ func TestShouldDispatchNotificationsToMultipleSubscribers(t *testing.T) {
 
 	l := logger.NewUPPLogger("test", "panic")
 	h := NewHistory(historySize)
-	d := NewDispatcher(delay, h, l)
+	e, err := access.CreateEvaluator(
+		"data.centralBanking.allow",
```
todor-videv1, line 91: This is a nitpick, in my implementation (in _notifications-rw_) I've strayed away from referencing Cenral Banking, preferring to use the more generic term special content. Maybe we should double check that with the product team?

todor-videv1:
Amazing job mate :)! Very neat and tidy!

todor-videv1 (approved):

app.go
ilian2233: It would be a good idea if the rule file was added as a config map. Given that the rule is basically a constant the package and variable names are also a constant. That's why I added it this way, but maybe it would be a good idea to export it as a constant.

access/evaluator.go
```diff
@@ -0,0 +1,33 @@
+package access
```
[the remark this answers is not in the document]
ilian2233: Yes, there will be other services with opa capabilities. The 3 services in the current tasks are more of a test to see if opa will work for us. Most likely in the future, the whole opa logic of all services will be exported in a separate service.

vlasakievft (dismissed):
🚀

todor-videv1 (approved):

martin-stanchev (approved):

PR #144, opened 2023-11-03
publication field awareness; small payload parsing refactoring

vlasakievft:
# Description

## What

Please be specific and try to describe your thought process. State the obvious, since this might be the first time the reviewer is looking at the code

## Why

[UPPSF-4694](https://financialtimes.atlassian.net/browse/UPPSF-4694)

## Anything, in particular, you'd like to highlight to reviewers

I made a small refactoring around payload parsing. We no longer using map[string]interface{} for json objects.

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Breaking change

## DoD - Ensure all relevant tasks are completed before marking this PR as "Ready for review"

- [ ] Test coverage is not significantly decreased
- [ ] All PR checks have passed
- [ ] Changes are deployed on dev before asking for review
- [ ] Documentations remains up-to-date
  - [ ] OpenAPI definition file is updated
  - [ ] README file is updated
  - [ ] Documentation is updated in upp-docs and upp-public-docs
  - [ ] Architecture diagrams are updated

___

[UPPSF-4694]: https://financialtimes.atlassian.net/browse/UPPSF-4694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


publication/publication.go
```diff
@@ -0,0 +1,74 @@
...
+	SustainableViews: "Sustainable Views",
+}
+
+type Publications struct {
+	UUIDS []uuid.UUID
+}
+
+// UnmarshslJSON
+func (p *Publications) UnmarshalJSON(b []byte) error {
+	uuids := strings.Split(string(b), "\"")
```
kalinmolerov, line 25: suggestion: standard Go unmarshal will work and it is much simpler:
err := json.Unmarshal(b, &publication);

publication/publication.go
```diff
@@ -0,0 +1,74 @@
...
+		if !v {
+			return p.String(), v
+		}
+	}
+	return "", v
+}
+
+// IsSunstainableViews return true if the content is generated or used
+// by the publication
+// Will return true in the case there is more than one publication supported
```
kalinmolerov, line 65: Code should not care if publication is SV or any other specialists title. For now we want to filter non Pink FT content. This is a very specialised case for SV. If another publication is added we have to make changes.

consumer/mapper.go
```diff
@@ -63,14 +55,17 @@ func (n NotificationMapper) MapNotification(event NotificationMessage, transacti
...
-		Title:            title,
+		EditorialDesk:    event.Payload.EditorialDesk,
+		Title:            event.Payload.Title,
 		SubscriptionType: contentType,
 	}
 
-	if n.IncludeScoop {
-		scoop := getScoopFromPayload(notificationPayloadMap)
-		notification.Standout = &dispatch.Standout{Scoop: scoop}
+	if event.Payload.Standout != nil {
```
kalinmolerov, line 63: I am not sure if this is correct. Scoop is part of Standout struct. There is no check if Scoop is true or false. Check is performed if Standout exists.
Note - Scoop is a very important field. It is monitored by our B2B clients. If we send by accident alerts for Scoop (not sure if this is the case) it will be a big problem!

opa_modules/special_content.rego
```diff
@@ -6,4 +6,8 @@ default allow := true
 
 allow = false if {
     input.EditorialDesk == "/FT/Professional/Central Banking"
-}
\ No newline at end of file
+}
+
+allow = false if {
+    input.IsSustainableViews == true
```
kalinmolerov, line 12: suggestion - change to input.NonFTPink == true. Will be more generic.

kalinmolerov (commented):
The implementation is too much based on Sustainable views. We want to stop all Non FT Pink publications from sending push notifications. I suggest to generalise the implementation and have clear separation between Ft Pink vs Non FT Pink without specifying Sustainable Views.

publication/publication.go
vlasakievft: This is publication package.

publication/publication.go
vlasakievft: This is not true.

dispatch/notifications.go
```diff
@@ -38,6 +38,7 @@ type NotificationModel struct {
 	Standout         *Standout
 	SubscriptionType string
 	IsE2ETest        bool
+	IsSV             bool
```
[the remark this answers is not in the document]
vlasakievft: We want policy to ignore SV content in case we want to enable sv content notifications.

consumer/mapper.go
vlasakievft: If Standout is always in the object, yes - we could return error if standout is missing.

ilian2233 (commented):
*suggestion:* On a general level, I would integrate the checks made in the `IsSustainableViews` function inside the rego file. That would solve most of Kalin's comments by making the implementation more generic and make use of `rego` to a fuller extend.

vlasakievft:
> _suggestion:_ On a general level, I would integrate the checks made in the `IsSustainableViews` function inside the rego file. That would solve most of Kalin's comments by making the implementation more generic and make use of `rego` to a fuller extend.
 
Thank you for the suggestion. In this case the implementation should not be generic. We support only two publications and in the event of open up the whole system to true multi-publication support, then we should think about how to handle this. I believe in this level of support, we are more transparent to ourself and the rest of the world.

opa_modules/special_content.rego
```diff
@@ -1,9 +1,18 @@
...
 
 allow = false if {
     input.EditorialDesk == "/FT/Professional/Central Banking"
-}
\ No newline at end of file
+}
+
+# Disable notifications for listed publications
+# Sustainable Views => 8e6c705e-1132-42a2-8db0-c295e29e8658
+block_notication_for_publication := ["8e6c705e-1132-42a2-8db0-c295e29e8658"]
```
kalinmolerov, line 14: Shouldn't the logic be opposite? Explicit allow instead of explicit deny?

consumer/mapper_test.go
```diff
@@ -1,20 +1,25 @@
...
 func TestMapToUpdateNotification(t *testing.T) {
 	t.Parallel()
 
-	standout := map[string]interface{}{"scoop": true}
-	payload := map[string]interface{}{"title": "This is a title", "standout": standout, "type": "Article", "publishCount": "2"}
+	itemJSON := `{"title":"This is a title", "standout": {"scoop": true}, "type": "Article", "publishCount": "2"}`
+
+	var payload Item
+
+	err := json.Unmarshal([]byte(itemJSON), &payload)
```
kalinmolerov, line 20: Why defining a json and then unmarshal it to an Item struct.  Why don't you directly define Item instance with values. I don't see how this unmarshal step is relevant to the test.

martin-stanchev:
Good work Vesko! One request: Please don't resolve comments on your own, since the commenter may have follow-up questions and they might not see your responses.

publication/publication.go
```diff
@@ -0,0 +1,86 @@
...
+func (p Publications) IsSustainableViews() bool {
```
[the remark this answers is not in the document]
vlasakievft: This is part of the publication package which I am trying to introduce. This method in this case we no longer use in this service.

publication/publication.go
vlasakievft: I use custom unmarshaling to encode uuids in proper way. See line 28

consumer/content.go
```diff
@@ -0,0 +1,23 @@
+package consumer
+
+import "github.com/Financial-Times/notifications-push/v5/publication"
+
+// Item struct represents the payload in NotificationMessage
+// We use omitempty to indicate nonmandatory fields
+type Item struct {
```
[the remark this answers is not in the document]
vlasakievft: This is a very good question. The reason I leave it that way is to engage in conversion how to proper structure this kind of "payload" objects. They should have a proper meaning to be more clearly understood.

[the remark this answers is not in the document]
martin-stanchev: You can see two examples on how we do this in other services here -> https://github.com/Financial-Times/enriched-content-rw-postgres/blob/main/schema/schema.go 
and here:
https://github.com/Financial-Times/post-publication-combiner/blob/master/processor/model.go

If there will be a lot of unmarshalling i think in general it's a good idea to have those in one place. But that's just my opinion so open to discussion here.

opa_modules/special_content.rego
kalinmolerov: After discussion with colleagues they think the logic should be Explicit Deny, so please ignore my previous comment.

consumer/content.go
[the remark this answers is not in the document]
vlasakievft: Exactly what I am looking for. In model.go I could see the whole picture of Content (our first citizen). Next step should be to move all common struct to one place. In this way we could have good documentation (automatically generated) and easy upgradability and many perks.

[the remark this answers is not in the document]
martin-stanchev: I think the main reason that there are separate in services is that they are mapped from different kafka messages. Post-publication-combiner gets the message from a different kafka topic, then modifies it in some ways and sends it to enriched-content-rw-postgres where it can have different fields, objects or their names in the json. This is also true for other services. We can think about exporting them, but only if we can be sure it's generic and reusable truly and not just the same, but exported to a different place.

dispatch/dispatcher.go
```diff
@@ -156,6 +156,32 @@ func (d *Dispatcher) forwardToSubscribers(notification NotificationModel) {
 		return
 	}
 
+	isPublicationAllowed := false
+	pub := notification.Publication
+	if pub != nil {
+		pu, pubErr := pub.OnlyOneAsString()
```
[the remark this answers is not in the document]
vlasakievft: This is temporary implementation and I believe the good behaviour is to stop articles with multiples publication.

publication/publication.go
```diff
@@ -0,0 +1,86 @@
+// Package publication repesents publication field in an Article or a Notification message
+package publication
+
+import (
+	"strings"
+
+	"github.com/google/uuid"
+)
+
+const PinkFt = "88fdde6c-2aa4-4f78-af02-9f680097cfd6"
```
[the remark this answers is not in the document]
vlasakievft: This is publication package and PinkFT is just a constant to represent the publication. I am not sure what do you mean.

vlasakievft:
> Good work Vesko! One request: Please don't resolve comments on your own, since the commenter may have follow-up questions and they might not see your responses.

Thank you, Martin! I would like to request every reviewer to actively resolve his/her comments if the goal is accomplished in a reasonable time. :  )

consumer/content.go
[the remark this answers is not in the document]
vlasakievft: And to answer to the file location of Item. It's clearer Item to be in separate file. And why is not named Payload I am a bit worried of heavy usage of Payload as a name. Is very generic and have no meaning. It's very difficult to understand what is happening behind the scene. Is payload a content of a type or something else? Questions like this. Item is not the correct name for sure.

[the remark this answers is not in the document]
martin-stanchev: Can you then extract the NotificationsMessage structs also to be in a separate file? It is not really clear when two relating structs are in separate places.

What do you think about NotificationItem, NotificationPayload, NotificationContents etc?

publication/publication.go
martin-stanchev: After having a second look in the publication package, the only method that is used outside of tests is  `OnlyOneAsString`, so what is this method needed for?

publication/publication.go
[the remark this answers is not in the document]
martin-stanchev: So can we remove it then?

dispatch/dispatcher.go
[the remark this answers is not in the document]
kalinmolerov: Didn't we agreed on the refinement on Thursday that we will have such case - multiple publications. So this pub.OnlyOneAsString() should be removed?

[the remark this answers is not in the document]
vlasakievft: Its outdated :  )

consumer/publish_event.go
```diff
@@ -61,11 +62,31 @@ type NotificationMessage struct {
...
+	Payload      Payload `json:"Payload,omitempty"`
 }
 
 // Matches is a method that returns True if the ContentURI of a publication event
 // matches a allowlist regexp
 func (e NotificationMessage) Matches(allowlist *regexp.Regexp) bool {
 	return allowlist.MatchString(e.ContentURI)
 }
+
+// Item struct represents the payload in NotificationMessage
```
martin-stanchev, line 74:
```suggestion
// Payload struct represents the payload in NotificationMessage
```

dispatch/dispatcher.go
```diff
@@ -156,6 +156,37 @@ func (d *Dispatcher) forwardToSubscribers(notification NotificationModel) {
 		return
 	}
 
+	isPublicationAllowed := false
+	pub := notification.Publication
+	if pub != nil {
+		pu, pubErr := pub.OnlyOneOrPink()
```
martin-stanchev, line 162: If we remove the code from line 162 until 170 and just pass the array of publications to OPA like you do in line 171, wouldn't the service function the exact same way it does now? OPA can match a string in the array and it would return if the publication is allowed or not if I am not mistaken.

publication/publication.go
kalinmolerov: Check this code in Go playground. I think unmarshal works with google uuid.
https://go.dev/play/p/IDJoJ2qwA5B

martin-stanchev (approved):

ilian2233 (approved):

kalinmolerov (approved):
I approve with special opinion. I don't think the model of content is correct - publication field. I don't like how some of the comments were addressed. We spend too much time on this task and we have more priority tasks. Fixing the tech debt added with this code should be addressed in the future task which should verify x-policy of the consumers.

